PR #77, opened 2018-08-02
FIX - Add `no-docker` argument to build command

dannypaz:
This PR changes the `npm run build` command to additionally build docker images. You can prevent docker-images from being built by using `npm run build no-docker`.


dannypaz:
@treygriffith you can view it here: https://github.com/sparkswap/docs/pull/6

PR #79, opened 2018-08-02
FIX - removed sample files and changed docker.for.mac to host.docker.internal

dannypaz:
This PR removes unused docker files (sample files) and changes our usage of `docker.for.mac.host.internal` to `host.docker.internal` due to the former's deprecation.

see: https://docs.docker.com/docker-for-mac/release-notes/#docker-community-edition-18030-ce-mac59-2018-03-26


dannypaz:
[networking set up on linux host machine](https://trello.com/c/kOnppj3f/585-networking-set-up-on-linux-host-machine)

PR #80, opened 2018-08-03
FIX - Add mainnet support for ltcd/btcd

dannypaz:
This PR allows the `$NETWORK` variable to be set correctly ONLY for testnet/simnet. When the `--network` option is excluded from btcd/ltcd startup, this allows ltcd/btcd to start on mainnet.


docker/ltcd/start-ltcd.sh
```diff
@@ -30,6 +29,11 @@ PARAMS=$(echo \
     "--txindex"
 )
 
+# If the network is mainnet, then we can omit the network tag for btcd.
```
treygriffith: ltcd

PR #81, opened 2018-08-03
FIX - added local setting for lnd-engine image build

dannypaz:
Currently, we default lnd-engine cert hosts to `host.docker.internal` which will only work for local development. This prevents users from having the correct certs when trying to test our the broker daemon according to the instructions on sparkswap.com.

This PR adds the default cert generation for ltcd/btcd to include `ltcd` and `btcd` as CNs, respectively. Additionally, this adds a new argument to the build command to allow the user to generate lnd-engine for local usage with `npm run build local`


treygriffith:
@dannypaz do we need to update the docs site anywhere to reflect these changes? or does this change make it fit with what the docs say to do ?

dannypaz:
@treygriffith These changes do not affect the docs, however they do affect building images with local use for the relayer.

This is fixing the issue where users had cert issues with the hosted version(edited). Our current default docker settings on the broker was assuming the user had a local relayer. Instead, the default setup is testnet.

PR #88, opened 2018-08-06
Feature/trading capacities

martineehrlich:
We need the local and remote trading capacities for all active, pending and inactive channels in order to create the trading statuses table for the CLI.


martineehrlich:
[Trading Status for Brokers](https://trello.com/c/phnPA11m/603-trading-status-for-brokers)

PR #89, opened 2018-08-09
Feature/close channel

martineehrlich:


src/engine-actions/close-channels.js
```diff
@@ -0,0 +1,83 @@
+const {
+  listChannels,
+  closeChannel,
+  listPendingChannels
+} = require('../lnd-actions')
```
dannypaz: [nit] add space

src/engine-actions/close-channels.js
```diff
@@ -0,0 +1,83 @@
...
+
+async function close (channel, client, logger) {
+  return new Promise((resolve, reject) => {
+    try {
+      const [fundingTxidStr, outputIndex] = channel.channelPoint.split(':')
+      const channelPoint = {
+        fundingTxidStr,
+        outputIndex: parseInt(outputIndex)
+      }
+      const close = closeChannel(channelPoint, false, { client })
```
dannypaz: We need to pass `force` for this method

src/engine-actions/close-channels.js
```diff
@@ -0,0 +1,83 @@
...
+ * @param {String} paymentChannelNetworkAddress
+ * @param {String} fundingAmount - int64 string
+ * @returns {Promise<void>} returns void on success
+ */
+async function closeChannels ({ force = false, pending = false } = {}) {
+  const { channels = [] } = await listChannels({ client: this.client })
+  const activeChannels = channels.filter(c => c.active)
+
+  let inactiveChannels = []
+  if (force === true) {
```
dannypaz: If specify force for `closeChannels` does this mean we only want to force close inactive channels? It sounds like what we actually want to do is `force` close ALL channels?

src/engine-actions/close-channels.js
```diff
@@ -0,0 +1,83 @@
...
+    inactiveChannels = channels.filter(c => c.active === false)
+  }
+
+  let pendingChannels = []
+  if (pending === true) {
+    const { pendingOpenChannels = [] } = await listPendingChannels({ client: this.client })
+    pendingChannels = pendingOpenChannels
+  }
```
dannypaz: At this point in the code, if all `activeChannels`, `inactiveChannels` or `pendingChannels` is of length 0, then we should return early and log.

src/engine-actions/close-channels.js
```diff
@@ -0,0 +1,83 @@
...
+        return finish(new Error(error))
+      }
+
+      const dataListener = (response) => {
+        logger.info('Closing channel', response)
+        return finish(null, response)
+      }
+
+      // Helper to make sure we tear down our listeners
+      const finish = (err, response) => {
```
dannypaz: [nit] although the ordering doesn't matter for this code to work, we should still always try to define things before we use them (line 56 uses this code)

src/engine-actions/close-channels.js
```diff
@@ -0,0 +1,83 @@
...
+      }
+      const close = closeChannel(channelPoint, false, { client })
+
+      const errorListener = (err) => {
+        logger.error(`Error from closeChannel stream`, err)
+        return finish(err)
+      }
+
+      const endListener = () => {
+        const error = `LND closed closeChannel stream before returning our value`
```
dannypaz: [nit] backticks should fail `standardjs` because we are not using interpolation. We also don't need this variable since the line length isn't that large for `logger.error`

src/lnd-actions/close-channel.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Open a channel w/ LND
```
dannypaz: `Close a channel`

src/lnd-actions/close-channel.js
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * @param {String} publicKey - lnd public key to open channel with
+ * @param {String} fundingAmount - the amount to fund the channel w/
```
dannypaz: fix the comments here

src/engine-actions/close-channels.js
```diff
@@ -0,0 +1,83 @@
...
+    inactiveChannels = channels.filter(c => c.active === false)
+  }
+
+  let pendingChannels = []
+  if (pending === true) {
+    const { pendingOpenChannels = [] } = await listPendingChannels({ client: this.client })
+    pendingChannels = pendingOpenChannels
+  }
+  const closedChannelResponses = await Promise.all(activeChannels.concat(inactiveChannels).map(channel => close(channel, this.client, this.logger)))
+  const closedPendingChannelResponses = await Promise.all(pendingChannels.map(channel => close(channel.channel, this.client, this.logger)))
```
dannypaz: For these 2 lines, instead of passing `this.channel` and `this.logger` we can use [`call`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call).

Example:
```
const closedPendingChannelResponses = await Promise.all(pendingChannels.map(channel => close.call(this, channel.channel)))
```

src/engine-actions/close-channels.js
martineehrlich: I'm going to leave the variable because it's used in two places.

src/engine-actions/close-channels.js
```diff
@@ -0,0 +1,70 @@
...
+  const activeChannels = channels.filter(c => c.active)
+
+  const closedChannelResponses = await Promise.all(activeChannels).map(channel => close(channel, force, this.client, this.logger))
+
+  this.logger.info('Successfully closed channels', closedChannelResponses)
+  return closedChannelResponses
+}
+
+async function close (channel, client, force, logger) {
+  logger.info('FORCE', force)
```
dannypaz: @martineehrlich did you still want this log? it looks like its a dev log

src/engine-actions/close-channels.js
```diff
@@ -0,0 +1,70 @@
...
+ * Closes a channel given the channel point
+ *
+ * @param {Boolean} [options.outbound=true] outbound is true if checking outbound channels, false if inbound
+ * @returns {Promise<Array.<Object>>}
+ */
+async function closeChannels ({ force = false } = {}) {
+  const { channels = [] } = await listChannels({ client: this.client })
+  const activeChannels = channels.filter(c => c.active)
+
+  const closedChannelResponses = await Promise.all(activeChannels).map(channel => close(channel, force, this.client, this.logger))
```
dannypaz: [nit] Instead of passing in client/logger to this function, we could use `close.call(this, channel, force)` which would do the same thing by binding the current context to the call.

src/lnd-actions/close-channel.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Close a channel w/ LND
+ * @see {@link http://api.lightning.community/#closechannel}
+ *
+ * @param {Object} channelPoint { fundingTxidStr, outputIndex } to identify the channel to be closed
+ * @param {Boolean} force - true if we want to force close the channel, false if not (defaults to false)
```
dannypaz: need to add client to these docs. 
```
@param {Object} options
@param {Object} options.client - lnd client
```

src/engine-actions/close-channels.js
```diff
@@ -0,0 +1,70 @@
+const {
+  listChannels,
+  closeChannel
+} = require('../lnd-actions')
+/**
+ * Closes a channel given the channel point
+ *
+ * @param {Boolean} [options.outbound=true] outbound is true if checking outbound channels, false if inbound
```
dannypaz: Need to change this to be `force`

src/engine-actions/close-channels.js
```diff
@@ -0,0 +1,70 @@
...
+ * @returns {Promise<Array.<Object>>}
+ */
+async function closeChannels ({ force = false } = {}) {
+  const { channels = [] } = await listChannels({ client: this.client })
```
dannypaz: If channels dont exist, we should log and return immediately. If we don't, we'll always receive `Successfully closed channels` even if there weren't any, which might be confusing

src/engine-actions/close-channels.js
martineehrlich: I'm using it for now haha, but I'll remove it when im done testing

src/engine-actions/close-channels.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  const closedChannelResponses = await Promise.all(activeChannels.map(channel => close(channel, force, this.client, this.logger)))
+
+  this.logger.info('Successfully closed channels', closedChannelResponses)
+  return closedChannelResponses
+}
+
+async function close (channel, force, client, logger) {
+  logger.info('FORCE', force)
```
dannypaz: Did you still want this log? @martineehrlich

dannypaz:
@martineehrlich changes look good. Please add a description to the PR and update w/ master

martineehrlich:
[wallet release command for broker](https://trello.com/c/2DQWfnO9/474-wallet-release-command-for-broker)

PR #91, opened 2018-08-15
add get max channel to engine actions

martineehrlich:
We want to know the max balance in the list of open channels so that we can decide on the broker if we need to open a new channel with a given balance, or if one already exists.

Add getMaxChannel method that takes in an outbound option to find the max of the local or the remote balance. 

https://github.com/sparkswap/broker/pull/235


src/engine-actions/get-max-channel.js
```diff
@@ -0,0 +1,22 @@
...
+async function getMaxChannel ({ outbound = true } = {}) {
+  const { channels = [] } = await listChannels({ client: this.client })
+  const balance = outbound ? 'localBalance' : 'remoteBalance'
+
+  if (channels.length === 0) {
+    this.logger.debug('getMaxChannel: No channels exist')
+    return {}
+  }
+
+  const maxBalance = channels.reduce((max, channel) => Big(channel[balance]).gte(max) ? Big(channel[balance]) : max, Big(channels[0][balance]))
```
dannypaz: [nit] Semantic change: since we are looking for max balance, we should use `gt` instead of `gte`

martineehrlich:
["Smart" commit](https://trello.com/c/9WwCPFJP/604-smart-commit)

src/engine-actions/get-max-channel.js
```diff
@@ -0,0 +1,22 @@
+const { Big } = require('../utils')
+const { listChannels } = require('../lnd-actions')
+
+/**
+ * Get local balance of all channels for a specific daemon
+ * @param {Boolean} [options.outbound=true] outbound is true if checking outbound channels, false if inbound
```
dannypaz: add the options hash param to jsdoc too
```
@param {Object} opts
```

dannypaz:
@martineehrlich https://github.com/sparkswap/lnd-engine/pull/91#pullrequestreview-146548976

PR #95, opened 2018-09-24
FIX - translateSwap routing error

dannypaz:
queryRoutes has the potential to return a failure if no target can be found (failure to find route in certain conditions). We want to handle this failure as a `permanent failure` instead of returning back to the caller.


dannypaz:
[Weird logs when interchain router fails for LND](https://trello.com/c/pf3wcaTM/747-weird-logs-when-interchain-router-fails-for-lnd)

src/engine-actions/translate-swap.js
```diff
@@ -88,10 +88,15 @@ async function translateSwap (takerAddress, swapHash, amount, extendedTimeLockDe
...
-    getInfo({ client: this.client })
-  ])
+  try {
+    var [ { routes = [] }, { blockHeight = '' } ] = await Promise.all([
+      queryRoutes(queryRoutesReq, { client: this.client }),
+      getInfo({ client: this.client })
+    ])
+  } catch (e) {
+    this.logger.error(e)
+    return { permanentError: e.message }
```
treygriffith: How can you be certain that the error returned is a permanent error? This could have a side effect of making the swap non-atomic in certain cases.

dannypaz: @treygriffith I think in this case, since we haven't started translation (only trying to find a route), that we are OK to send a permanent failure. FWIW, we return a permanent error is `routes = 0`, so if this call fails, it is technically the same thing.

treygriffith: Yes, I agree, since this is just pulling information and not initiating a payment, it is ok to do a permanent error here.

PR #98, opened 2018-10-11
WIP: convert from common to quantum when adding an invoice

treygriffith:
When creating an invoice on the engine, the engine should be responsible for conversion from the common units (e.g. BTC) to the quantum units (e.g. Satoshis).

This is a breaking change in this API for all consumers.


treygriffith:
Closing this since we don't have a case yet where we need to create an invoice from common units.

PR #101, opened 2018-10-17
Fix - Add pgp verification to bitcoind

dannypaz:
This PR adds additional security to bitcoind download in our bitcoind docker files. Previously, we had used a trusted apt-get repo (ppa) to install bitcoin. We now download the binary from bitcoin.org and check to make sure the binary was signed by a trusted maintainer.

This emulates the setup for litecoind.

Additionally, i've bumped the version of bitcoind to 0.17.x


dannypaz:
[Add gpg verification for bitcoind](https://trello.com/c/yeL74MvD/808-add-gpg-verification-for-bitcoind)

PR #102, opened 2018-10-22
FIX - Fix mining for bitcoind/litecoind

dannypaz:
This PR fixes mining for bitcoind/litecoind where the cron process could not grab the correct binary from /usr/local.

You can view if miner is running by looking at the logs in `/jobs/cron.log`


docker/bitcoind/start-bitcoind.sh
```diff
@@ -44,16 +44,37 @@ if [[ "$DEBUG" == "debug" ]]; then
...
+
+# For a regtest setup, we need to expose a port that other nodes can connect to.
+# this port will be designated for ONLY miners.
+if [[ "$NETWORK" == "regtest" ]] && [[ "$MINER" = true ]]; then
+    PARAMS="$PARAMS -port=$MINER_PORT"
+fi
 
 # If MAX_WEB_SOCKETS is specified then we'll set it in the params, otherwise
 # we will let bitcoind decide on the default (125)
+# NOTE: This should only be used for regtest environments
```
treygriffith: why not make this an explicit condition?

docker/lnd/start-lnd-ltc.sh
```diff
@@ -35,7 +35,7 @@ PARAMS=$(echo \
     "--extpreimage.rpchost=$EXTPREIMAGE_HOST" \
     "--$NODE.rpcuser=$RPC_USER" \
     "--$NODE.rpcpass=$RPC_PASS" \
-    "--$NODE.rpchost=$RPC_HOST"
+    "--$NODE.rpchost=$RPC_HOST" \
```
treygriffith: [nit] why put the trailing slash on the last line?

PR #105, opened 2018-11-05
WIP Feature/remove wallet

dannypaz:
This PR is the beginning of multiple changes to allow the lnd-engine to support password and cipher seed backups for lnd wallets.

PR #106, opened 2018-11-05
Feature - WalletUnlocker Initialization

dannypaz:
**NOTE** This PR can be merged as is. It is backwards compatible with the current setup. #105 will be the last PR to be merged

This PR builds ontop of https://github.com/sparkswap/lnd-engine/pull/105 to allow use of the `WalletUnlocker` RPC from LND.

When LND has been started for the first time, no wallet or macaroons are present. LND will also disable access to its `Lightning` RPC until a wallet is created.

In this interim stage, a `WalletUnlocker` service is exposed that allows the user to create/init/unlock a particular wallet.

Once the wallet is initialized/unlocked, LND then removes the `WalletUnlocker` service and the `Lightning` RPC is now fully functional.

In order to support this, we've refactored the `generate-lnd-client` methods into 2 separate modules to handle the above situation.

**NOTE** Renaming `this.client` is out of the scope of this ticket, as it would provide a massive `diff`. I will try to rename that specifically in another PR.


dannypaz:
[Support `WalletUnlocker` in lnd-engine](https://trello.com/c/gAuNO11J/854-support-walletunlocker-in-lnd-engine)

src/lnd-setup/generate-wallet-unlocker-client.js
```diff
@@ -0,0 +1,40 @@
...
+    throw new Error(`LND-ENGINE error - tls cert file not found at path: ${tlsCertPath}`)
+  }
+
+  const tls = fs.readFileSync(tlsCertPath)
+  const tlsCredentials = grpc.credentials.createSsl(tls)
+
+  try {
+    return new lnrpc.WalletUnlocker(host, tlsCredentials)
+  } catch (e) {
+    logger.error('Unable to create WalletUnlocker')
```
treygriffith: what's the scenario in which we can't generate this client but we still want to be able to operate? I understand for the Lightning Client, but not the WalletUnlocker - shouldn't it always be available?

src/utils/load-proto.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * Load Proto
+ * @module src/utils/load-proto
+ */
+
+const grpc = require('grpc')
+const grpcProtoLoader = require('@grpc/proto-loader')
```
treygriffith: Does this upgrade have to go with this PR? It seems pretty unrelated and the PR is pretty large as it is.

src/lnd-setup/generate-lightning-client.js
```diff
@@ -0,0 +1,66 @@
...
+
+    const sslCredentials = grpc.credentials.createSsl(tls)
+    rpcCredentials = grpc.credentials.combineChannelCredentials(sslCredentials, macaroonCredentials)
+  }
+
+  try {
+    return new lnrpc.Lightning(host, rpcCredentials, {})
+  } catch (e) {
+    logger.error('Unable to create Lightning rpc', e)
+    return null
```
treygriffith: This seems like a fairly blunt instrument - there are myriad reasons why creating a Lightning client would fail, but it will always get interpreted as "not unlocked". Is there no other way to detect the state of this client?

src/index.js
```diff
@@ -36,10 +47,19 @@ class LndEngine {
...
 
     Object.assign(this, actions)
   }
+
+  /**
+   * Client will return false if the lnd-engine requires wallet configuration before
+   * the engine is fully functional.
+   */
+  get isUnlocked () {
+    return !!Object.keys(this.client).length
```
treygriffith: This seems like a pretty strange way to detect the unlocked status.

A more understandable way to do this (if it's possible) would be to do something like try the client and update a specific property (e.g. `isUnlocked`) if it responds with a particular error code.

src/lnd-setup/generate-wallet-unlocker-client.js
dannypaz: They are interchangable. If the `Lightning` client is not avail, then WalletUnlocker is, but the RPC goes away once the wallet has been unlocked

src/utils/load-proto.js
dannypaz: @treygriffith the upgrade was already merged to `lnd-engine`. I just moved the loading of the proto file out into its own method

src/index.js
dannypaz: I wish it gave us a specific error code. We'll have to think about this. What happens is all of our `client` methods become undefined. GRPC will return a blank object for `Lightning` which is why we check for the length of keys (which would be 0 and false)

src/lnd-setup/generate-lightning-client.js
dannypaz: @treygriffith maybe we change `isUnlocked` naming, but either way, we log that we were unable to create the rpc w/ its error. I have no found any RPCs or statuses that would give us this information, unfortunately

src/lnd-setup/generate-wallet-unlocker-client.js
treygriffith: It seems like you should always be able to build these clients from the proto file even if the service is down, so this seems like a bad way to detect which service is being used.

src/utils/load-proto.js
treygriffith: I see - deleted file made the diff hard to parse.

src/lnd-setup/generate-lightning-client.js
treygriffith: Seems strange since you're loading the client from a file. Let's investigate to see if we can nail down the specific behavior.

src/lnd-setup/generate-wallet-unlocker-client.js
dannypaz: Sorry Trey, I was really wrong on this implementation. During development, there were a few issues I had that led me to this error, one of them being: missing deconstruction of `{ lnrpc }` which led to the handling of this error AND the previous implementation of `isUnlocked`. Please check out the latest commits for changes.

src/lnd-setup/generate-lightning-client.js
dannypaz: @treygriffith after investigation, the current functionality is: calling `new lnrpc.Lightning` works and if we try to make a service call, we will receive a `code 12` unimplemented error. Please check out recent commits for changes.

src/index.js
dannypaz: @treygriffith this has been fixed in recent commits. Thank you for pushing on a more thorough investigation

src/utils/load-proto.js
dannypaz: Yup, that was my bad. Deleted the `generate-lnd-client` file and created 2 more

src/lnd-actions/gen-seed.js
```diff
@@ -0,0 +1,20 @@
+const { deadline } = require('../grpc-utils')
+
+/**
+ * Generates a mnuemonic seed used to recover a user's wallet
+ *
+ * @function
+ * @see {@link http://api.lightning.community/#genSeed}
+ * @return {Object} res
+ * @return {Object} cipherSeedMnemonic
```
treygriffith: dual return value is weird here, as is having the return value be anything other than a Promise

src/lnd-setup/generate-lightning-client.js
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @function
+ * @param {LndEngine} engine
+ * @param {String} engine.host
+ * @param {String} engine.protoPath
+ * @param {String} engine.tlsCertPath
+ * @param {String} engine.macaroonPath
+ * @param {Logger} engine.logger
+ * @returns {grpc.Client} lnrpc Lightning client definition
+ * @returns {null}
```
treygriffith: this looks like an extra return note

src/lnd-setup/generate-wallet-unlocker-client.js
```diff
@@ -0,0 +1,35 @@
...
+ * Generates a lnrpc.WalletUnlocker client which is only used on initialization of the LND
+ * node.
+ *
+ * @function
+ * @param {LndEngine} - engine
+ * @param {String} engine.host
+ * @param {String} engine.protoPath
+ * @param {String} engine.tlsCertPath
+ * @returns {grpc.Client} lnrpc WalletUnlocker client definition
+ * @returns {null}
```
treygriffith: extra return value?

src/lnd-setup/generate-wallet-unlocker-client.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+function generateWalletUnlockerClient ({ host, protoPath, tlsCertPath }) {
```
treygriffith: this feels pretty similar to the other client - any way to refactor some of it (without losing legibility)?

src/engine-actions/validate-node-config.js
```diff
@@ -1,13 +1,78 @@
...
+  // and validate if the engine is available or not
+  this.isUnlocked = isUnlocked
+
+  // If the Lightning service is not available (wallet is not unlocked) on the engine
+  // then we should check the only non-modifying endpoint of the WalletUnlocker to make
+  // sure that a connection to the daemon can be made.
+  if (!isUnlocked) {
+    try {
+      await genSeed({ client: this.walletUnlocker })
+      return true
```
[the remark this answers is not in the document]
dannypaz: @treygriffith One thing we will have to deal with is... we cannot validate the configuration of the node _until_ the wallet has been unlocked. `getInfo` rpc only exists on `Lightning` service. We need a check to make sure the RPC is available in the method.

I dont have a good answer for this, but will think about the implementation

src/engine-actions/validate-node-config.js
```diff
@@ -1,13 +1,78 @@
...
+ * @throws {Error} LND has no chains configured
+ * @throws {Error} LND can only support one active chain at a time
+ * @throws {Error} Mismatched configuration if chain is different than engine configuration
  */
 async function validateNodeConfig () {
+  const isUnlocked = await isEngineUnlocked(this.client)
+
+  // We need to set the `isUnlocked` variable here so that end users can call the engine
+  // and validate if the engine is available or not
+  this.isUnlocked = isUnlocked
```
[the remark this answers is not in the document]
dannypaz: This is what we will need to do, because validateNodeConfig can only run on a node that has a wallet (more info https://github.com/sparkswap/lnd-engine/pull/106#discussion_r231765013)

src/lnd-actions/gen-seed.js
dannypaz: This is a jsdoc mistake.

src/lnd-setup/generate-wallet-unlocker-client.js
dannypaz: Since the implementation of `generate-wallet-unlocker` is pretty straight-forward, it would probably be more code to tear some of it out. I made the decision with this piece of code as a "times when it would be good to have duplication".

Having either the `WalletUnlocker` or `Lightning` services be returned from the same function, I found that all I was doing was adding in conditionals.

With this said, if you have suggestions for refactoring, let me know. I can re-evaluate this approach

src/engine-actions/validate-node.js
```diff
@@ -0,0 +1,78 @@
...
+    // unimplemented.
+    // In GRPC Unimplemented indicates operation is not implemented or not
+    // supported/enabled in this specific service.
+    if (e.code && e.code === UNIMPLEMENTED_SERVICE_CODE) {
+      return false
+    }
+
+    // We return true, because the error is unrelated to a service not being implemented
+    // and the user will now have to troubleshoot further
+    return true
```
treygriffith: It seems weird to return true if we don't know whether the engine is unlocked or not. Maybe it's just semantics, but it feels like this is more of an unknown state.

src/engine-actions/validate-node.js
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Validates this engine
+ *
+ * @function
+ * @return {True} Node is configured correctly
+ * @throws {Error} Failed to validate an unlocked engine
+ */
+async function validateNode () {
+  const isUnlocked = await isEngineUnlocked(this.client)
```
treygriffith: why not set this directly to `this` instead of having an intermediate value?

src/engine-actions/validate-node.js
dannypaz: @treygriffith I was worried about generic errors from LND. An example is LND throwing an error `block out of order`... `getInfo` would fail but the engine is technically unlocked.

dannypaz: @treygriffith in the recent commits, we always return `FALSE` unless we can prove that the `Lightning` RPC is up.

src/engine-actions/is-node-config-valid.js
```diff
@@ -5,10 +5,19 @@ const { getInfo } = require('../lnd-actions')
...
- * @return {void}
+ * @throws {Error} Not Unlocked
+ * @throws {Error} No Chains Configured
+ * @throws {Error} Only allow, at most, one active chain
+ * @throws {Error} Mismatched Configuration
+ * @return {Boolean}
  */
-async function validateNodeConfig () {
-  const { chains = [] } = await getInfo({ client: this.client })
+async function isNodeConfigValid () {
```
dannypaz: Renamed `validateNodeConfig` to `isNodeConfigValid` to fit in-line with our naming scheme for other engine actions.

src/engine-actions/is-node-config-valid.js
```diff
@@ -24,7 +33,7 @@ async function validateNodeConfig () {
     throw new Error(`Mismatched configuration: Engine is configured for ${this.currencyConfig.chainName}, LND is configured for ${chainName}.`)
   }
 
-  this.validated = true
+  return true
```
dannypaz: Removed assignment of `this.validated` to the LndEngine constructor for readability, instead of setting different engine properties nested in other functions

src/index.js
```diff
@@ -67,18 +88,32 @@ class LndEngine {
...
-      // We do not await this function because we want the validations to run in the background.
-      // It can take time for the engines to be ready, so we use exponential backoff to retry validation
-      // for a period of time, until it is either successful or there is actually something wrong.
-      const validationCall = () => this.validateNodeConfig()
       const payload = { symbol: this.symbol }
       const errorMessage = 'Engine failed to validate. Retrying'
+      const validationCall = async () => {
+        // We make a initial call to check if engine is unlocked. The result updates
+        // `this.unlocked` and is then used in the underlying configuration check.
+        this.unlocked = await this.isEngineUnlocked()
```
treygriffith: I don't think we need to save the state of the `unlocked` status since we don't use it outside of this call and instead depend on the engine's overall validation status.

dannypaz: @treygriffith as a note, this may be needed, but can be added in a later PR. The broker _could_ call `isEngineUnlocked` separately, but we would probably want to "cache" that information instead of continually making an RPC call

PR #109, opened 2018-11-09
Feature/add validation check wrapper

martineehrlich:
For most engine actions we need to check that the engine is validated before they are called now that we are allowing users to use the broker without engine validation. 

This adds a wrapper around engine actions needing validation that adds a check to see if the engine is valid (otherwise throws an error). Also adds a validated flag that gets set to true once validateNodeConfig succeeds. 

https://github.com/sparkswap/broker/pull/272


martineehrlich:
[Wait for LND to be ready before trying to start sparkswapd](https://trello.com/c/cKzRhlyM/608-wait-for-lnd-to-be-ready-before-trying-to-start-sparkswapd)

src/engine-actions/validate-node-config.js
```diff
@@ -24,7 +24,7 @@ async function validateNodeConfig () {
     throw new Error(`Mismatched configuration: Engine is configured for ${this.currencyConfig.chainName}, LND is configured for ${chainName}.`)
   }
 
-  return true
+  this.validated = true
```
treygriffith: make sure to update the jsdocs

PR #112, opened 2018-11-16
Feature: SendToRoute using a stream

treygriffith:
Instead of using `sendToRouteSync`, we replace it with `sendToRoute` and read the stream directly. This provides more flexibility for a potentially long-running HTLC settlement process.


src/engine-actions/translate-swap.js
```diff
@@ -44,6 +44,84 @@ function calculateTimeLock (extendedTimeLockDelta, secondsPerBlock, blockHeight)
...
+ * @param  {grpc.Client} options.client
+ * @return {Promise<TranslateSwapOutcome>}
+ */
+function sendToRouteSync (paymentHash, routes, { logger, client }) {
+  return new Promise((resolve, reject) => {
+    try {
+      logger.debug('Setting up stream to sendToRoute', { paymentHash })
+      const stream = sendToRoute({ client })
+
+      // keep local state of whether we cancelled so that we can
```
dannypaz: [nit] whether we manually cancelled the stream?

src/engine-actions/translate-swap.js
```diff
@@ -44,6 +44,84 @@ function calculateTimeLock (extendedTimeLockDelta, secondsPerBlock, blockHeight)
...
+ * gRPC error code for a CANCELLED request, either by a client or a server
+ * @see {@link https://github.com/grpc/grpc/blob/master/doc/statuscodes.md}
+ * @constant
+ * @type {Number}
+ */
+const CANCELLED_CODE = 1
+
+/**
+ * Use the sendToRoute function as though it returned a promise
+ * @param  {String}      paymentHash     Base64 string of the payment hash to pay for
```
dannypaz: Shouldn't this be `bytes` instead? looks like the call to sendToRoute either accepts paymentHash or paymentHashString

src/lnd-actions/send-to-route.js
```diff
@@ -1,33 +1,12 @@
...
  */
-function sendToRoute (paymentHash, routes, { client }) {
-  return new Promise((resolve, reject) => {
-    /** Although the LND RPC proto suggest that it supports a `bytes` parameter of `paymentHash`
-     * It does not pass through correctly when using `SendToRouteSync` and rejects as an invalid payment hash
-     * @see {@link https://github.com/lightningnetwork/lnd/blob/master/rpcserver.go#L2356}
-     * @see {@link https://trello.com/c/lc3URJ4G/335-submit-bugfix-for-lnd-sendtoroutesync-paymenthash}
-     */
-    try {
-      const paymentHashString = Buffer.from(paymentHash, 'base64').toString('hex')
```
dannypaz: Depending on the type we choose for `paymentHash`, we might need this code in the engine action

src/engine-actions/translate-swap.js
```diff
@@ -44,6 +44,84 @@ function calculateTimeLock (extendedTimeLockDelta, secondsPerBlock, blockHeight)
   return maxTimeLockDeltaInBlocks.plus(blockHeight).toString()
 }
 
+/**
+ * gRPC error code for a CANCELLED request, either by a client or a server
+ * @see {@link https://github.com/grpc/grpc/blob/master/doc/statuscodes.md}
+ * @constant
+ * @type {Number}
+ */
+const CANCELLED_CODE = 1
```
dannypaz: Very british of you

src/engine-actions/translate-swap.js
treygriffith: our settings on grpc convert Base64 strings to `bytes`, so the description is accurate.

src/engine-actions/translate-swap.js
treygriffith: Eh, I think the "manual" isn't really the distinction, it's just whether it was cancelled by the client or the server.

src/engine-actions/translate-swap.js
treygriffith: 🇬🇧

PR #116, opened 2018-11-26
FIX - bitcoind/litecoind regtest fund generation

dannypaz:
Currently, we mine 101 blocks every minute to however this floods our daemons with transactions which gets to be cumbersome and confusing. 

This PR removes the 101 block initialization and places this block requirement on manual generation from the parent (the service that owns the bitcoind/litecoind regtest nodes)


dannypaz:
Tabling this ticket for now

PR #117, opened 2018-11-27
Feature - Unlocked state in lnd-engine

dannypaz:
To provide better feedback to the broker regarding the state of an lnd-engine, we set an `unlocked` property which is then used to give feedback during validation dependent actions.


dannypaz:
[[EPIC] Add status check to let the user know they need to unlock wallet](https://trello.com/c/iuBFXBuz/860-epic-add-status-check-to-let-the-user-know-they-need-to-unlock-wallet)

src/index.js
```diff
@@ -92,9 +100,9 @@ class LndEngine {
         //
         // We make this initial call to check if engine is unlocked before continuing
         // to validate the current engine
-        const engineIsUnlocked = await this.isEngineUnlocked()
+        this.unlocked = await this.isEngineUnlocked()
```
martineehrlich: Nit, `isEngineUnlocked` is a bit redundant.

dannypaz: This change is actually outside the scope of this feature. `isEngineUnlocked` was added in a previous PR. Once we create a standard for the engine (arch design) we can revisit (ive added a card to our backlog JIC)

martineehrlich: I agree that this change is outside the scope of the PR, but I don't understand how `Once we create a standard for the engine (arch design)` is relevant to the redundant naming?

It seems like the pattern we follow is not to have engine in the name of functions? ex. `isAvailable`, `isNodeConfigValid` so `isEngineUnlocked` goes against that pattern.

dannypaz: We'll eventually need to publish a spec for an engine, which at that time we can modify the calls. I understand that `engine` isn't in the name of any other function, but its beside the point. I agree that the name can be changed

PR #118, opened 2018-11-28
Feature - Enable wallets by default

dannypaz:
This PR removes the `noseedbackup` flag from LND as the default. This flag had previously allowed LND to initialize a wallet on behalf of the user that was not recoverable OR password protected.

By removing this flag, the user will now have to create a wallet on LND engine startup and each subsequent restart of the engine, will lock the lnd engine requiring a password.

In previous PRs we have supplied a `createWallet` and `unlockWallet` API which is used for both of these actions.

Additionally, this PR looks for a `NO_SEED_BACKUP` environment variable that enables the `--noseedbackup` flag for LND *ONLY* if the engine is set to something other than MainNet (real money)

### DIFF
--- a/docker/lnd/start-lnd-btc.sh
+++ b/docker/lnd/start-lnd-btc.sh
@@ -50,7 +50,7 @@ fi
 # We want to make it easy for devs to test functionality of the engine which could
 # potentially involve the constant restarting of a particular daemon. It can get
 # annoying to continually have to unlock/create wallets so we will enable noseedbackup
-# ONLY if the network is outside of mainnet AND dev is set to true
+# ONLY if the network is outside of mainnet AND `NO_SEED_BACKUP` is set to true
 if [[ "$NO_SEED_BACKUP" == true ]] && [[ "$NETWORK" != 'mainnet' ]]; then
     PARAMS="$PARAMS --noseedbackup"
     echo 'LND has --noseedbackup set. You are at your own peril'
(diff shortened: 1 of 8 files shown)


dannypaz:
[[EPIC] Remove `--no-seedbackup` configuration](https://trello.com/c/1VoOmUsh/853-epic-remove-no-seedbackup-configuration)

docker/lnd/start-lnd-btc.sh
```diff
@@ -51,4 +47,13 @@ if [[ "$NODE" == "bitcoind" ]]; then
     PARAMS="$PARAMS --bitcoind.zmqpubrawtx=$ZMQPUBRAWTX"
 fi
 
+# We want to make it easy for devs to test functionality of the engine which could
+# potentially involve the constant restarting of a particular daemon. It can get
+# annoying to continually have to unlock/create wallets so we will enable noseedbackup
+# ONLY if the network is outside of mainnet AND dev is set to true
```
treygriffith: your comment mentions that this only happens if dev is set to true, but the conditional doesn't check that. Which one is wrong?

dannypaz: the comment is wrong. I changed the name from DEV to `NO_SEED_BACKUP`

dannypaz:
Closing this PR. We can reopen when sparkswap is ready to fully lock down the engine's

PR #121, opened 2018-11-30
exclude spec and test folders from coverage check

dannypaz:
We want to exclude spec files and test folders from the coverage check in lnd-engine. We do not require any files in `lnd-actions` to be tested.


package.json
```diff
@@ -12,6 +12,13 @@
       "docs/*"
     ]
   },
+  "nyc": {
+    "exclude": [
+      "**/*.spec.js",
+      "src/lnd-actions/*",
+      "test/"
```
treygriffith: test should be excluded by default

package.json
```diff
@@ -12,6 +12,13 @@
       "docs/*"
     ]
   },
+  "nyc": {
+    "exclude": [
+      "**/*.spec.js",
+      "src/lnd-actions/*",
```
treygriffith: not sure I agree with excluding this, although I agree writing tests for this directory has not been a priority for us.

dannypaz: OK. I will keep it simple

dannypaz: This might be an OK addition as I think if we start rejecting PR on coverage than this will be a factor. I dont see us adding specs to those file as they are mainly a way for us to promisify the lnd calls

PR #123, opened 2018-12-05
Feature: get payment preimage

treygriffith:
### Description
This change adds two new engine actions:
- isPaymentPendingOrComplete
- getPaymentPreimage

They can be used together to retrieve a payment's preimage if it has already been paid, but return otherwise. This is *not* the same as idempotency for `sendPayment`, which will still send a payment if one has not previously been sent for this hash, which is not always desired behavior.

### Example Usage
```
if (await engine.isPaymentPendingOrComplete(paymentHash)) {
  return engine.getPaymentPreimage(paymentHash)
} else {
  // do something else, like send a payment
}
```

### Related Pull Requests
Relies on changes in https://github.com/sparkswap/lnd/pull/22

### Notes
Based on changes expected to be made in upstream LND to make `SendPayment` idempotent and potentially to add streaming payment status as an RPC, these engine actions may change in their implementation, but their signature to the engine's consumer should be able to stay fairly consistent.

### Todo
- [ ] tests


src/engine-actions/is-payment-started.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+const PAYMENT_STATUSES = Object.freeze({
+  GROUNDED: 'GROUNED',
+  IN_FLIGHT: 'IN_FLIGHT',
+  COMPLETE: 'COMPLETE'
+})
+
+/**
+ * Retrieves whether a given payment (as indicated by its hash) has
+ * started. Payments that are currently in-progress or completed are
```
martineehrlich: It's a little confusing that a complete payment is categorized as a started payment

src/engine-actions/get-payment-preimage.js
```diff
@@ -0,0 +1,86 @@
...
+ * showing up in this list, which will mistakenly throw an error.
+ * {@link https://github.com/lightningnetwork/lnd/issues/2032}
+ */
+async function getCompletedPreimage (paymentHash, { logger, client }) {
+  const payments = await listPayments({ client })
+
+  logger.debug(`Found ${payments.length} completed payments`)
+
+  const paymentHashBuf = Buffer.from(paymentHash, 'base64')
+  const payment = payments.find(({ paymentHash }) => {
```
martineehrlich: What is happening here? Can you add a comment?

src/engine-actions/is-payment-started.js
```diff
@@ -0,0 +1,41 @@
...
+ * considered to have started, whereas payments that have failed
+ * or have never been started are considered not to have started.
+ *
+ * @function
+ * @param  {String} paymentHash Base64 encoded payment hash
+ * @return {Boolean} true if a payment is in progress or completed; false otherwise
+ */
+async function isPaymentStarted (paymentHash) {
+  const client = this.client
```
dannypaz: destructure this you savage

src/engine-actions/get-payment-preimage.js
```diff
@@ -0,0 +1,79 @@
...
+ * @param  {String} paymentHash Base64 encoded payment hash
+ * @return {String} Base64 encoded payment preimage associated with the hash
+ * @throws {Error} If payment is not in-flight or completed
+ */
+async function getPaymentPreimage (paymentHash) {
+  const { client, logger } = this
+  const status = await lookupPaymentStatus(paymentHash, { client })
+
+  if (status === PAYMENT_STATUSES.GROUNDED) {
+    throw new Error(`No payment with hash '${paymentHash}'' is in-flight or complete`)
```
martineehrlich: I think there is an extra apostrophe here

src/engine-actions/get-payment-preimage.js
```diff
@@ -0,0 +1,79 @@
...
+ * @param  {Object} options
+ * @param  {Object} options.client
+ * @return {String} Base64 encoded payment preimage associated with the hash
+ * @throws {Error} If payment is not completed
+ * @todo LND has a known issue preventing payments that are settled after a restart
+ * showing up in this list, which will mistakenly throw an error.
+ * {@link https://github.com/lightningnetwork/lnd/issues/2032}
+ */
+async function getCompletedPreimage (paymentHash, { logger, client }) {
+  const payments = await listPayments({ client })
```
martineehrlich: You need to destructure payments out of listPayments, because listPayments returns an object with the list inside, not the list itself.

src/engine-actions/get-payment-preimage.js
```diff
@@ -0,0 +1,81 @@
...
+ * @param  {Object} options
+ * @param  {Object} options.client
+ * @return {String} Base64 encoded payment preimage associated with the hash
+ * @throws {Error} If payment is not completed
+ * @todo LND has a known issue preventing payments that are settled after a restart
+ * showing up in this list, which will mistakenly throw an error.
+ * {@link https://github.com/lightningnetwork/lnd/issues/2032}
+ */
+async function getCompletedPreimage (paymentHash, { logger, client }) {
+  const { payments } = await listPayments({ client })
```
martineehrlich: What happens if there are no payments? I think you need to default payments to an empty array.

PR #125, opened 2018-12-06
FIX - Try to close inactive channels without force

dannypaz:
We want to try to close all open channels and have `closeChannels` error if we have failed. Currently, we filter out inactive channels, but this is incorrect because we would like to know whether we should force close channels or not.

This PR removes the `active` filter from channels


dannypaz:
[Bug: CloseChannels not attempting to close inactive channels](https://trello.com/c/4iiXz0gl/941-bug-closechannels-not-attempting-to-close-inactive-channels)

PR #127, opened 2018-12-10
added fee estimation to create-channel

dannypaz:
We currently have commit limits to BTC and LTC when a "user" uses the broker. When a user tries to commit their currency under these limits, there are times when the fees are too high and do not cover the opening transaction

This PR adds a very archaic "fee estimation" that provides a buffer for the engine so that we can commit all of their available currency with respect to fees.

We've set conservative, hardcoded limits for each currency. In the future we will expose fee estimation through the engine to provide realtime limits.


src/engine-actions/create-channel.js
```diff
@@ -14,17 +18,54 @@ const {
...
+  // If the funding amount and fees are greater than our current uncommitted balance
+  // AND the funding amount cannot cover the fees, then we should error to the user
+  // because we would fail to fund a channel. This also prevents against having negative
+  // values for a fundingAmount after a fee estimate has been taken
+  if (balanceCantCoverFees && Big(feeEstimate).gt(fundingAmount)) {
+    throw new Error('fundingAmount does not cover estimated fees', { fundingAmount, feeEstimate })
+  } else if (balanceCantCoverFees) {
+    this.logger.debug('Current engine balance does not have enough funds for channel and fees', { fundingAmount, feeEstimate, balance })
+
+    const fundingAmountSubFees = Big(fundingAmount).minus(feeEstimate).toString()
```
treygriffith: [nit] I would call this `fundingAmountLessFees`

treygriffith:
@dannypaz can you add `madge` to this PR for cyclic dependency detection?

PR #129, opened 2018-12-11
Bugfix/include pending channels in max channel

martineehrlich:
When committing funds, we check if there is already a channel open for that currency. We were having an issue where we were allowing multiple channels of the same currency to open if you committed balances very quickly in a row. This was because the committed balance of the first channel was still pending.

This PR also takes pending channels into consideration when attempting to open new channels.


martineehrlich:
[When committing a balance, behavior should be the same regardless of amount in channel](https://trello.com/c/HmKWvwtT/948-when-committing-a-balance-behavior-should-be-the-same-regardless-of-amount-in-channel)

src/engine-actions/get-max-channel.js
```diff
@@ -8,16 +8,42 @@ const { listChannels } = require('../lnd-actions')
...
+  ] = await Promise.all([
+    // The response from listChannels consists of channels that may be active or inactive
+    listChannels({ client: this.client }),
+    listPendingChannels({ client: this.client })
+  ])
   const balance = outbound ? 'localBalance' : 'remoteBalance'
 
-  if (channels.length === 0) {
-    this.logger.debug('getMaxChannel: No channels exist')
+  if (channels.length === 0 && pendingOpenChannels.length === 0) {
```
dannypaz: [nit] A recent trick that I learned was `if (!channel.length && !pendingOpenChannels.length)`, since having a length of `0` will evaluate to false for JS

src/engine-actions/get-max-channel.js
```diff
@@ -8,16 +8,42 @@ const { listChannels } = require('../lnd-actions')
...
+    this.logger.debug('getMaxChannel: No open or pending channels exist')
     return {}
   }
 
-  const maxBalance = channels.reduce((max, channel) => Big(channel[balance]).gt(max) ? Big(channel[balance]) : max, Big(channels[0][balance]))
-  return { maxBalance: maxBalance.toString() }
+  const maxBalance = channels.reduce((max, channel) => {
+    if (Big(channel[balance]).gt(max)) {
```
dannypaz: [nit] not apart of your PR, but using `balanceType` instead of `balance` would make this code more readable

src/engine-actions/get-max-channel.js
```diff
@@ -8,16 +8,42 @@ const { listChannels } = require('../lnd-actions')
...
+      return Big(channel[balance])
+    } else {
+      return max
+    }
+  }, Big('0'))
+
+  const maxPendingBalance = pendingOpenChannels.reduce((max, pendingChannel) => {
+    if (Big(pendingChannel.channel[balance]).gt(max)) {
```
dannypaz: Something we might want to cleanup is that IF we used a map on `pendingOpenChannels` so that the structure was the same as `channels`, we could break out this reduce function into its own method `findMaxChannel` (since this call is `getMaxChannel`) and use it in both places. It could potentially be easier to test

PR #131, opened 2018-12-13
Add gpg pool for litecoind/bitcoind gpg verification

dannypaz:
This PR updates the bitcoind/litecoind docker keyserver url from `pgp.mit.edu` to `pool.sks-keyservers.net` to provide more stability when building lnd-engine.

Additionally, I've refactored the litecoind `Dockerfile` to match bitcoind

**Issue:**
We've had a few cases in the past couple of weeks where `pgp.mit.edu` has failed to verify pgp keys and returned the following information:

```
gpg: keyserver receive failed: No data
```

This exits the build process for lnd-engine and prevents users from using the latest version of the code.

**Technical Background:**

We (crypto community as a whole) use pgp keys to verify the contents of a bitcoind or litecoind release to ensure that the code has not been tampered with


dannypaz:
[Bug: if MIT keyserver fails, then build fails for lnd-engine](https://trello.com/c/lF4pBP3k/977-bug-if-mit-keyserver-fails-then-build-fails-for-lnd-engine)

PR #134, opened 2018-12-13
removed node patch version from node-version

dannypaz:


dannypaz:
[Update node versions in lnd-engine](https://trello.com/c/QZWfgR6x/929-update-node-versions-in-lnd-engine)

PR #136, opened 2018-12-14
Fix - successful validation after validation retries

dannypaz:
Stick with me cause this is going to be a long one.

In order to identify if an LND instance (engine) is in a certain state, we use the `genSeed` method. (simply cause there is no identifier on lnd to let us know what kind of state its in).

This poses issues where LND may or may not have access to `genSeed` and may or may not have access to a macaroon file.

This PR contains changes to `isEngineUnlocked` and `validateNodeConfig` code that allows validation to occur, as expected, in the following circumstances:

1. An engine is locked when a wallet needs to be created
2. An engine is locked when a wallet is available, but needs a password to be unlocked
3. An engine is unlocked when a wallet was created AFTER validation had failed
4. An engine is unlocked when a wallet is available AFTER validation had failed

This PR resolves #3 and #4, where the engine would act like it was in a `locked` even after commands had been made to unlock a wallet successfully. Once the lnrpc activates on lnd, it creates a macaroon, but we never regenerate `this.client` which causes failures until the broker is restarted (through docker)

### DIFF
--- a/src/utils/exponential-backoff.js
+++ b/src/utils/exponential-backoff.js
@@ -48,10 +48,8 @@ async function exponentialBackoff (callFunction, payload = {}, { errorMessage = 
       const nextDelayTime = delayTime * DELAY_MULTIPLIER
 
       if (errorMessage) {
-        logger.debug(errorMessage, { payload, delayTime, attemptsLeft, error: error.message, stack: error.stack })
         logger.error(errorMessage, { payload, delayTime, attemptsLeft, error: error.message })
       } else {
-        logger.error(`Error calling ${callFunction}. Retrying in ${Math.round(delayTime / 1000)} seconds, attempts left: ${attemptsLeft}`, { payload, error: error.message, stack: error.stack })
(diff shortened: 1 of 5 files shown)


dannypaz:
[Node being down shows as "locked"](https://trello.com/c/F4rJK0gx/936-node-being-down-shows-as-locked)

src/utils/exponential-backoff.js
```diff
@@ -48,9 +48,9 @@ async function exponentialBackoff (callFunction, payload = {}, { errorMessage =
       const nextDelayTime = delayTime * DELAY_MULTIPLIER
 
       if (errorMessage) {
-        logger.error(errorMessage, { payload, delayTime, attemptsLeft, error })
+        logger.error(errorMessage, { payload, delayTime, attemptsLeft, error: error.message })
```
dannypaz: Depending on the error, we may never receive the correct text. Adding `error.message` provides more info in the logs for exponential backoff.

dannypaz:
@treygriffith consolidating this into https://github.com/sparkswap/lnd-engine/pull/137 because im removing half of the files in this PR

PR #137, opened 2018-12-15
Add `available` status to engine

dannypaz:
This PR adds an additional `available` state to the engine so that we can determine whether an engine is available, locked, or unlocked.

- An engine is available if we can make requests to the RPC server without timeouts
- An engine is locked if we need to create a wallet or unlock a wallet
- An engine is unlocked if it is fully ready to take requests on the Lightning Network (lnrpc)

We ran into an issue where the Engine would tell the consumer that 'Engine is locked' when the engine was actually down or unavailable (we wouldnt be able to know if the engine was locked in a down or unavailable state).

Ive modified the `isAvailable` method for 2 reasons:
1. The method forced the consumer to wrap `isAvailable` in a try-catch to know if the engine was _actually_ available
2. This seemed like the best method to perform this check, outside of creating other methods like `isEngineAvailable` which we wanted to stay away from


treygriffith:
I don't think I like this approach, although I admit it may be the only way. I'm going to attempt to lay out my understanding of the states and we can discuss further.

LND State                      | Indicators
-----------------------|-----------
Unavailable / Unknown | lnrpc `UNAVAILABLE`, wallet unlocker rpc `UNAVAILABLE`
Wallet does not exist    | lnrpc `UNIMPLEMENTED`, `genSeeds` does not throw
Wallet is locked             | lnrpc `UNIMPLEMENTED`, `genSeeds` throws "wallet exists" (err code?)
Unlocked                        | lnrpc `getInfo` does not throw
Validated config            | lnrpc `getInfo` matches our configuration

I think whatever we do, we should have a unified view of the above table (or a modified version of it, if it is incorrect).

Maybe that means having a single command that returns the state as an enum instead of a series of booleans? I think I would like that better than having the caller interpret these outcomes (as is happening in https://github.com/sparkswap/broker/pull/328).

And given that you have added a requirement for handling when a node goes down while the engine process is still running, perhaps we also need to reset this state when we encounter an error in the engine action wrappers.

Happy to discuss further in person as well.

dannypaz:
@treygriffith Updated the table slightly. Thank you for writing it out. I totally agree with what you are saying. There may be an approach where we can return a `STATUS` about the engine, while also providing the helpers `available`, `locked`, etc. I will try to clean it up.

RE: https://github.com/sparkswap/broker/pull/328 My approach is wrong on that PR, however I would expect that the broker would map statuses into a 'user friendly' format, so there would still be instances in broker code where we would want to use that pattern (if you know what i mean)

src/engine-actions/get-status.js
```diff
@@ -0,0 +1,67 @@
...
+      .catch((e) => {
+        if (e.code !== grpc.status.UNIMPLEMENTED) {
+          this.logger.error('LndEngine failed to call getInfo', { error: e.message })
+          return resolve(ENGINE_STATUS.UNAVAILABLE)
+        }
+
+        genSeed({ client: this.walletUnlocker })
+          .then(() => resolve(ENGINE_STATUS.NEEDS_WALLET))
+          .catch((e) => {
+            if (e.message && e.message.includes(WALLET_EXISTS_ERROR_MESSAGE)) {
```
treygriffith: is there no code for this error? Do we have to do string matching?

src/utils/engine-status.js
```diff
@@ -0,0 +1,24 @@
...
+ * NEEDS_WALLET - Wallet does not exist,	lnrpc UNIMPLEMENTED, genSeeds does not throw
+ * LOCKED - lnrpc UNIMPLEMENTED, genSeeds throws "wallet exists"
+ * UNLOCKED - lnrpc getInfo does not throw
+ * VALIDATED - lnrpc getInfo matches our configuration
+ *
+ * @constant
+ * @type {Object<key, String>}
+ * @default
+ */
+const ENGINE_STATUS = Object.freeze({
```
treygriffith: Not a fan of this being defined in utils. Perhaps config, or even at the top level?

src/index.js
```diff
@@ -93,31 +84,29 @@ class LndEngine {
...
-          throw new Error('LndEngine is locked, unable to validate config')
-        }
+        // Returns a status `ENGINE_STATUS`
+        this.status = await this.getStatus()
 
-        // Once the engine is unlocked, we will attempt to validate our engine's
-        // configuration. If we call `isNodeConfigValid` before the engine is unlocked
-        // the call will fail without a friendly error
-        this.validated = await this.isNodeConfigValid()
+        if (!this.validated) {
```
treygriffith: where is `this.validated` being set?

src/index.js
```diff
@@ -93,31 +84,29 @@ class LndEngine {
...
+        // on LND. Since an engine can become unlocked during any validation call, we
+        // need to ensure that we are updating LndEngine's client in the situation that
+        // a macaroon file becomes available (meaning an engine has been unlocked).
+        this.client = generateLightningClient(this)
 
-        if (!this.unlocked) {
-          throw new Error('LndEngine is locked, unable to validate config')
-        }
+        // Returns a status `ENGINE_STATUS`
+        this.status = await this.getStatus()
```
treygriffith: do we expect external users to access `this.status` ? If not, does it make sense to provide users some way to get the current status of the engine without making gRPC calls? (Perhaps this is something we can punt as well)

src/utils/exponential-backoff.js
```diff
@@ -33,28 +33,28 @@ const EXPONENTIAL_BACKOFF_DELAY = 5000
...
-async function exponentialBackoff (callFunction, payload = {}, { errorMessage = null, attempts = EXPONENTIAL_BACKOFF_ATTEMPTS, delayTime = EXPONENTIAL_BACKOFF_DELAY, logger = console }) {
+async function exponentialBackoff (callFunction, payload = {}, { debugName = null, attempts = EXPONENTIAL_BACKOFF_ATTEMPTS, delayTime = EXPONENTIAL_BACKOFF_DELAY, logger = console }) {
   try {
     var res = await callFunction()
   } catch (error) {
     if (attempts > 0) {
       const attemptsLeft = attempts - 1
       const nextDelayTime = delayTime * DELAY_MULTIPLIER
 
-      if (errorMessage) {
```
treygriffith: Why remove this functionality?

src/index.js
dannypaz: forgot to checkin the `getter` for this property

src/utils/engine-status.js
dannypaz: @treygriffith top level works for me

src/engine-actions/get-status.js
dannypaz: Not sure if my comment showed up. This returns a code 2 which is pretty generic (we usually see 4s or 2s. I can check for the code also, but string matching had to suffice (I had a comment for doing this in a previous PR)

src/index.js
dannypaz: To get the status without grpc calls, they can use `engine.status`, but we they would have to call `engine.getStatus()` then `this.status` (which I dont dig, but is OK for now). Does that work?

src/utils/exponential-backoff.js
dannypaz: @treygriffith that was the only place we used it (and didnt need it)

src/index.js
treygriffith: Yes we can punt more advanced usage for later

src/engine-actions/get-status.spec.js
```diff
@@ -0,0 +1,112 @@
+const grpc = require('grpc')
+const path = require('path')
+const { expect, rewire, sinon } = require('test/test-helper')
+
+const LndEngine = rewire(path.resolve(__dirname, '..', 'index'))
+const getStatus = rewire(path.resolve(__dirname, 'get-status'))
+
+describe.only('get-status', () => {
```
treygriffith: hanging `only`

src/engine-actions/get-status.js
```diff
@@ -0,0 +1,77 @@
...
+ */
+const WALLET_EXISTS_ERROR_MESSAGE = 'wallet already exists'
+
+/**
+ * Returns the state of the current lnd-engine.
+ * @see {LndEngine#ENGINE_STATUS}
+ * @returns {String} status - ENGINE_STATUS
+ */
+function getStatus () {
+  return new Promise((resolve, reject) => {
```
treygriffith: why implement this as a Promise instead of `async`? And same question for each subsequent call - I think `await` with a try/catch wrapper is more readable than this

src/index.spec.js
```diff
@@ -134,41 +138,53 @@ describe('lnd-engine index', () => {
       expect(logger.error).to.have.been.calledWith(sinon.match('Failed to validate engine'))
     })
 
+    describe('get validated', () => {
```
treygriffith: you should add a test for the positive case

treygriffith: and you should actually test the `validated` property

src/engine-actions/get-status.js
dannypaz: @treygriffith multiple try catches were super ugly before, but ill try it again and see if it clears anything up

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "lnd-engine",
-  "version": "0.2.3-alpha-preview",
-  "description": "A Sparkswap Engine for LND lightning implementation",
+  "version": "0.3.0-beta-rc1",
+  "description": "A SparkSwap Engine for LND lightning implementation",
```
treygriffith: we're using `Sparkswap` now

src/engine-actions/get-status.js
```diff
@@ -0,0 +1,76 @@
...
+  getInfo,
+  genSeed
+} = require('../lnd-actions')
+
+/**
+ * @constant
+ * @type {String}
+ * @default
+ */
+const WALLET_EXISTS_ERROR_MESSAGE = 'wallet already exists'
```
treygriffith: I would add some description about what this is, including a link to the LND docs or code where it is defined.

src/index.js
```diff
@@ -16,6 +18,29 @@ const { exponentialBackoff } = require('./utils')
  */
 const LND_PROTO_FILE_PATH = require.resolve('../proto/lnd-rpc.proto')
 
+/**
+ * The current states of an lnd-engine
+ *
+ * UNKNOWN - Default state
```
treygriffith: [nit] Maybe inline these comments?

src/index.js
```diff
@@ -51,31 +76,20 @@ class LndEngine {
...
-    // passed through the constructor of the Engine.
-    //
-    // The configuration of the Engine lets the user know what currencies/chains are
-    // currently supported, as well as providing assurance that communication to
-    // an engine's node is available.
-    //
-    // We set `validated` to false by default, however this will be modified in the
-    // `validateEngine` action
-    this.validated = false
+    // Default status of the lnd-engine is unknown as we have not ran any validations
```
treygriffith: `run`, not `ran` here

src/index.js
```diff
@@ -51,31 +76,20 @@ class LndEngine {
...
+    // up to this point
+    this.status = ENGINE_STATUS.UNKNOWN
+    this.statuses = ENGINE_STATUS
```
treygriffith: should we export these on the constructor instead of the instance?

PR #139, opened 2018-12-18
remove noseed backup flag except for in simnet

martineehrlich:
We do not want to have a seed backup in non simnet environments. 

1) Remove the flag for regtest, testnet, mainnet. Users will have to enter a password when creating a wallet now.
2) Keep the flag for simnet so we can just use the default.


docker/lnd/start-lnd-btc.sh
```diff
@@ -51,4 +51,8 @@ if [[ "$NODE" == "bitcoind" ]]; then
     PARAMS="$PARAMS --bitcoind.zmqpubrawtx=$ZMQPUBRAWTX"
 fi
 
+if [[ "$NETWORK" == 'simnet' ]]; then
+  PARAMS="$PARAMS ----noseedbackup=1"
```
dannypaz: change the backup string to `--noseedbackup`

martineehrlich:
[Remove `--no-seedbackup` configuration](https://trello.com/c/1VoOmUsh/853-remove-no-seedbackup-configuration)

docker/lnd/conf/lnd-btcd.conf
```diff
@@ -14,9 +14,6 @@ debuglevel=info
 maxpendingchannels=2
 nobootstrap=1
 
-; TODO: Change to allow the user to encrypt wallet for MAINNET
-noseedbackup=1
-
```
dannypaz: [nit] can we not delete this space. My OCD is triggered

PR #140, opened 2018-12-19
Fix lookup payment status and bump version

treygriffith:
This change fixes a reference to an enum returned from LND, and since it was successfully tested, subsequently version-bumps the package to beta-rc1.


dannypaz:
@treygriffith a previous PR added `beta-rc1` so you might have conflicts? but this LGTM

PR #141, opened 2018-12-19
noseedbackup param does not need an arg in command line

martineehrlich:
When starting lnd_ltc and lnd_btc on simnet, we were getting an error where noseedbackup had an unexpected argument. This came from moving the noseedbackup flag from the conf file to the bash file. In the conf file the noseedbackup needed to be set to 1 where in the command line, it is just a flag (no argument).


martineehrlich:
[bugfix, noseedbackup should not have argument](https://trello.com/c/FjPIi389/1001-bugfix-noseedbackup-should-not-have-argument)

dannypaz:
@martineehrlich this was only occuring with LTC?

PR #148, opened 2019-01-11
Add max channel size and max payment

treygriffith:
This PR adds max channel size and max payment size to the currency config for the engine.

It also adds a simple check for max channel size when opening a channel.


src/config.json
```diff
@@ -6,15 +6,19 @@
       "symbol": "BTC",
       "quantumsPerCommon": "100000000",
       "secondsPerBlock": 600,
-      "feeEstimate": "20000"
+      "feeEstimate": "20000",
+      "maxChannelBalance": "16777215",
+      "maxPaymentSize": "4194304"
```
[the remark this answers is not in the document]
treygriffith: @dannypaz no this is a good catch. I thought it was 1/4 of the max channel size, but it looks like it's ~1/4 of the max channel size.

src/config.json
```diff
@@ -6,15 +6,19 @@
       "symbol": "BTC",
       "quantumsPerCommon": "100000000",
       "secondsPerBlock": 600,
-      "feeEstimate": "20000"
+      "feeEstimate": "20000",
+      "maxChannelBalance": "16777215",
+      "maxPaymentSize": "4294967"
```
martineehrlich: Where did you get these values? Can we document somewhere so we know where to find them later

dannypaz: @treygriffith could probably just throw a little section in the README under the heading `Config`

src/index.js
```diff
@@ -19,6 +19,21 @@ const {
  */
 const LND_PROTO_FILE_PATH = require.resolve('../proto/lnd-rpc.proto')
 
+/**
+ * Config properties that should be exposed directly on the engine
+ * @constant
+ * @type {Array<String>}
+ */
+const PUBLIC_CONFIG = [
+  'symbol',
```
dannypaz: if we are adding `symbol` here, I would remove the line above that sets `this.symbol` (this would override it)

src/index.js
```diff
@@ -47,6 +62,18 @@ class LndEngine {
       throw new Error(`${symbol} is not a valid symbol for this engine.`)
     }
 
+    // Expose config publicly that we expect to be used by consumers
+    PUBLIC_CONFIG.forEach((configKey) => {
```
dannypaz: Im cool with this, but what was the reasoning for choosing this over simple `get` getters? It'd be more readable just to include only what values you need with `get maxChannelBalance` etc..

Configurations that we need are `secondsPerBlock`, `maxChannelBalance`, `maxPaymentSize`. If the rest of the properties are meant to be private, then we can simply use `currencyConfig` from inside the engine?

src/engine-actions/translate-swap.js
```diff
@@ -144,13 +144,7 @@ async function translateSwap (takerAddress, swapHash, amount, extendedTimeLockDe
...
   const { publicKey } = networkAddressFormatter.parse(takerAddress)
 
-  const secondsPerBlock = this.currencyConfig.secondsPerBlock
-
-  if (!secondsPerBlock) {
-    const err = 'secondsPerBlock is not specified in the currencyConfig for lnd-engine'
-    this.logger.error(err)
-    return { permanentError: err }
-  }
+  const secondsPerBlock = this.secondsPerBlock
```
dannypaz: since this is internal, i'm still cool with using `this.currencyConfig.secondsPerBlock` unless we need to expose this to the user

src/engine-actions/get-status.js
```diff
@@ -44,8 +44,8 @@ async function getStatus () {
 
     const [ chainName ] = chains
 
-    if (chainName !== this.currencyConfig.chainName) {
-      this.logger.error(`Mismatched configuration: Engine is configured for ${this.currencyConfig.chainName}, LND is configured for ${chainName}.`)
+    if (chainName !== this.chainName) {
```
dannypaz: same here https://github.com/sparkswap/lnd-engine/pull/148/files#r247667603

src/index.js
dannypaz: @treygriffith I might opt for removing it, since we set the symbol in the constructor (and already check if its available)

dannypaz:
@treygriffith other than the last comment, I dig these changes.

src/config.json
dannypaz: @treygriffith are you cool with these comments or do you think it makes it messy?

src/index.js
```diff
@@ -47,6 +61,14 @@ class LndEngine {
...
+      if (this.currencyConfig[configKey] == null) {
```
dannypaz: OK so JavaScript is weird with this one because of the coercion (`undefined == null`), can we use `this.currencyConfig.hasOwnProperty(configKey)` instead. We can then have values of null or 0 or whatever for the config files.

dannypaz: I assumed we probably used this line to have a positive instead of negative statement*

treygriffith: Yeah I did this to catch `undefined` and `null`, but still allow `0` and other falsey values. The double equals was intentional. But I think that `hasOwnProperty` is cleaner, I'll do that.

PR #151, opened 2019-01-17
Improve locked status and timeout

treygriffith:
This PR is two changes to improve user feedback when unlocking a wallet.

- adds a new getter on the engine, `isLocked`, that returns whether the engine is currently in a locked state.
- adds a standard timeout to wallet unlocks to prevent them from hanging forever

This will allow the Broker to provide better user feedback when unlocking the engine's wallets.


dannypaz:
@treygriffith can you version bump in this PR?

PR #153, opened 2019-01-18
Remove SimNet

dannypaz:


dannypaz:
[Deprecate simnet](https://trello.com/c/q0Amviwn/1051-deprecate-simnet)

PR #156, opened 2019-02-05
Fix - LND Cert generation

dannypaz:
This PR makes changes to the lnd-engine docker setup to have LND auto generate certs. Previously, we were unable to use LND to generate certs because they used a p-curve that was unsupported by NodeJS. This has since been changed to use p-256 (on LND) which allows us to remove custom cert generation from the dockerfile.

**Changes**
1. remove cert generation from Dockerfile (now relies on LND)


scripts/build-images.sh
```diff
@@ -20,13 +20,13 @@ COMMIT_SHA=${LND_VERSION}
 LND_BTC_NODE=bitcoind
 LND_LTC_NODE=litecoind
 
-# NOTE: The names specified with `-t` directly map to the service names in
-# the applicable services docker-compose file
-docker build -t sparkswap_lnd_btc ./docker/lnd --build-arg NODE=$LND_BTC_NODE --build-arg NETWORK=btc --build-arg COMMIT_SHA=$COMMIT_SHA
-docker build -t sparkswap_lnd_ltc ./docker/lnd --build-arg NODE=$LND_LTC_NODE --build-arg NETWORK=ltc --build-arg COMMIT_SHA=$COMMIT_SHA
+# NOTE: The names specified with `-t` directly map to the our service names in
+# on sparkswap dockerhub
+docker build -t sparkswap/lnd_btc:latest ./docker/lnd --build-arg NODE=$LND_BTC_NODE --build-arg NETWORK=btc --build-arg COMMIT_SHA=$COMMIT_SHA
```
treygriffith: can we do this before we reference these names in the docker-compose file?

scripts/build-images.sh
```diff
@@ -20,13 +20,13 @@ COMMIT_SHA=${LND_VERSION}
...
+docker build -t sparkswap/lnd_ltc:latest ./docker/lnd --build-arg NODE=$LND_LTC_NODE --build-arg NETWORK=ltc --build-arg COMMIT_SHA=$COMMIT_SHA
```
treygriffith: Should we tag this with `latest` by default? Or tag it with the current version from package.json? Or the commit hash?

Is this how we publish images as well, or is this just building, with publishing requiring a separate `tag` step?

dannypaz: @treygriffith we're going to have to rethink our docker setup but here is why we tag latest:

- by default, if we omit the tag it goes to latest
- our docker-compose files use latest (which is easier *for us* than having them as a specific version)

By having all images tagged to latest, `latest` can then mean `latest sparkswap version` OR `the version I build on my machine`

With this said, the more and more we lock these version (including locking them in broker/relayer docker-compose), the harder it will be to dev so we'll have to take that into consideration)

TBH we're probably better off tagging the version from package.json or something.

RE: your last comment, publishing does require a separate tag step (so we can technically have a different name locally than on dockerhub)

scripts/build-images.sh
dannypaz: @treygriffith can you elaborate? Do you mean building this image in broker instead?

scripts/build-images.sh
dannypaz: @treygriffith TBH the thing im running into is that even though the code for lnd-engine may change, that doesnt mean the image has changed.

It feels weird to continually tag an image `v0.5.0`, `v0.5.1` when the images have had no changes

dannypaz:
@treygriffith I thought a bit about the setup yesterday and landed on this:

For every image in the lnd-engine (`lnd_btc`, `lnd_ltc`, `litecoind`, `bitcoind`), we will tag these to the version listed in the package.json

For our applications, we will peg the versions of these images in the docker-compose file, and using the `regtest.sample` we will be able to have fluid versions for development.

Although it might be pointless to keep tagging image versions even though there were *no* image changes, it is worthwhile for us to do so (so that we can have correct versioning for the entire application stack).

We can revisit this if it becomes and issue

dannypaz:
[Publish Docker Images (New Release)](https://trello.com/c/Jwjl1ju5/1102-publish-docker-images-new-release)

PR #162, opened 2019-02-22
Fix - Regtest Funding

dannypaz:
This PR adds functionality to where `101` blocks is mined ONLY ONCE at the beginning of either `bitcoind` or `litecoind`'s life during regtest where `MINER` is set to true.

1 block will be mined every 10 seconds
On initial startup 101 blocks will be mined for segwit


docker/litecoind/start-litecoind.sh
```diff
@@ -23,7 +23,7 @@ if [[ "$NETWORK" == "regtest" ]] && [[ "$MINER" = true ]]; then
     # `cron` instead of `crond`, where the former does not import environment variables
     # from the user session and instead uses a bare-bones `sh` session to run these commands
     LITECOIN_CLI_PATH=/usr/local/bin/litecoin-cli
-    echo "* * * * * ( $LITECOIN_CLI_PATH -regtest -rpcuser=$RPC_USER -rpcpassword=$RPC_PASS generate 101 >> /jobs/cron.log 2>&1 ) ; sleep infinity" >> /jobs/funding-cron.txt
+    echo "* * * * * (  ) ; sleep infinity" >> /jobs/funding-cron.txt
```
treygriffith: ?

PR #166, opened 2019-03-09
Feature/get max channel for address

martineehrlich:
We need an engine action for getting max channel for a given address.

1) Refactor getting channels with a given address.
2) Add get max channel for address
3) Have num channels for address and max channel for address use the helper function.


src/engine-actions/get-channels-for-address.js
```diff
@@ -0,0 +1,34 @@
...
+  ] = await Promise.all([
+    // The response from listChannels consists of channels that may be active or inactive
+    listChannels({ client: this.client }),
+    listPendingChannels({ client: this.client })
+  ])
+
+  const normalizedPendingChannels = pendingOpenChannels.map(chan => chan.channel)
+
+  if (channels.length === 0 && normalizedPendingChannels.length === 0) {
+    this.logger.debug('getChannelsForAddress: No channels exist')
```
dannypaz: @martineehrlich should we return here instead of going through the rest of the code? (not sure if we do this in the other files)

src/engine-actions/get-channels-for-address.js
```diff
@@ -0,0 +1,34 @@
+const { networkAddressFormatter } = require('../utils')
+const { listChannels, listPendingChannels } = require('../lnd-actions')
+
+/**
+ * Get all channels to the given address
+ * @param {string} address
+ * @returns {Array} channels
+ */
+async function getChannelsForAddress (address) {
```
dannypaz: The assumption in this method is that we are searching for a remote address. Is it overkill to have this be `getChannelsForRemoteAddress`?

src/engine-actions/get-channels-for-address.js
dannypaz: Looks like we do return early for `getMaxChannel`. We should return early here with `return {}`

dannypaz:
+1 on the quick changes @martineehrlich theres some good clean up in here

martineehrlich:
[Relayer checks all channels instead of scoping to single broker when trying to commit funds](https://trello.com/c/Ox09itqE/1229-relayer-checks-all-channels-instead-of-scoping-to-single-broker-when-trying-to-commit-funds)

src/engine-actions/get-max-channel-for-address.js
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Get maximum balance from all channels (inbound or outbound) for a given address
+ * @param {string} address
+ * @param {Object} [options={}]
+ * @param {boolean} [options.outbound=true] - outbound is true if checking outbound channels, false if inbound
+ * @returns {Object} res
+ * @returns {string} res.maxBalance - the max balance in all open channels.
+ */
+async function getMaxChannelForAddress (address, { outbound = true } = {}) {
```
dannypaz: `getMaxChannelBalanceForAddress` and only return the balance in this method. Since its only one value for this function, we dont need to return an object.

It looks like the relayer doesn't have any other uses for the old method `getMaxChannel` so we should be OK to change the API for this new function

dannypaz: I retract my comment. @martineehrlich recommended we wait for the change and that sounds right. The reason why `getMaxChannelForAddress` and `getMaxChannel` returns an empty object, is so that if NO channels exist, we'll receive an undefined value in the code.

dannypaz:
@martineehrlich ty for adding docs

PR #169, opened 2019-03-15
Fix - docker apt package cache

dannypaz:
When we run `apt-get update` in a Dockerfile (or any similar command), we need to make sure that it is not using the cache. This PR adds a fix to our docker files so that `apt` cache is busted when it is updated

See approach here: http://lenguyenthedat.com/docker-cache/


dannypaz:
[Bust cache when running `apt-get update`](https://trello.com/c/CFCan00H/1235-bust-cache-when-running-apt-get-update)

PR #170, opened 2019-03-18
Feature - Specify Fees for LND

dannypaz:
This PR allows the user to set the `BASE_FEE` and `FEE_RATE` environment variables on their LND instance to handle what type of routing fees they want to allow


dannypaz:
[Remove Relayer routing fees](https://trello.com/c/kI2DkJOO/1245-remove-relayer-routing-fees)

PR #173, opened 2019-03-23
Bugfix/double counting uncommitted pending

martineehrlich:
When closing a channel, we were getting a duplicated pending unconfirmed balance. This is because unconfirmedBalance for the wallet and waitingCloseChannels contained the same information. We can remove the waitingCloseChannels for this reason, and we can also remove the pendingClosingChannels because it deprecated (https://github.com/lightningnetwork/lnd/blob/15c9b389fa097b6beb0b0e8f24ba78a9d1254693/rpcserver.go#L2167).

Now the actual unconfirmed balance is the combination of what is the unconfirmedBalance in the wallet and the balance in pending force close channels.


src/engine-actions/get-uncommitted-pending-balance.js
```diff
@@ -7,15 +7,14 @@ const { Big } = require('../utils')
...
-  if (pendingClosingChannels.length === 0 && pendingForceClosingChannels.length === 0 && waitingCloseChannels.length === 0) {
-    this.logger.debug('getUncommittedPendingBalance: No channels exist')
+  if (!pendingForceClosingChannels.length) {
+    this.logger.debug('getUncommittedPendingBalance: No pendingForceClosingChannels exist')
     return Big(unconfirmedBalance).toString()
   }
 
-  const allPendingChannels = pendingClosingChannels.concat(pendingForceClosingChannels).concat(waitingCloseChannels)
-  const totalPendingLocalBalance = allPendingChannels.reduce((acc, c) => {
+  const totalPendingLocalBalance = pendingForceClosingChannels.reduce((acc, c) => {
```
treygriffith: I would add some commentary here about the values that we are ignoring, and why (they are zero, or they are already included in `unconfirmedBalance`

PR #175, opened 2019-03-25
Fix - refactor build images

dannypaz:
We have some left over code from our old `npm run build local` commands from lnd-engine. Additionally, the command `npm run build-images` is rarely used as `npm run build` is the general command that we have across repos.

This PR consolidates our build scripts in the engine for a single command `npm run build`


dannypaz:
[Add initiator field to channel lists](https://trello.com/c/U60KjmaZ/1337-add-initiator-field-to-channel-lists)

PR #178, opened 2019-03-26
extend connect peer deadline to receive lnd error

martineehrlich:


martineehrlich:
Closing this because we are going to attempt to make a connection with the relayer for both currencies which should make it so the relayer is already connected to the broker by the time this call  is being made. (We should not have a ports issue at this point)

PR #180, opened 2019-03-28
replace exponential backoff with perpetual retry

treygriffith:
Exponential backoff is a good idea in theory, but causes lots of user experience issues.

To name a few:
- after blocks have been syncing for hours, it takes a very long time before the status changes to "synced" because of the backoff
- if you unlock wallets after having the broker up for a few minutes, the status does not update very quickly

So, we remove exponential backoff and replace it with a much simpler retry of 10 seconds.


dannypaz:
[Remove exponential backoff for LND health checks](https://trello.com/c/WINB3wNM/1357-remove-exponential-backoff-for-lnd-health-checks)

PR #181, opened 2019-03-29
make EXTERNAL_ADDRESS optional for lnd setup

dannypaz:
`EXTERNAL_ADDRESS` should be an optional property on our LND setup, as a static IP is not required to use the relayer/lightning network


dannypaz:

PR #182, opened 2019-03-29
FEATURE - Connect user

dannypaz:
This PR adds a `connectUser` API endpoint so that Broker's can have a way to connect to the relayer without needing to expose external IP addresses from their host machine. This allows us to make EXTERNAL_ADDRESS optional for all current engines


dannypaz:

PR #183, opened 2019-03-29
removed host flag and stopped erroring if no host existed

dannypaz:
This PR modifies `getPaymentChannelNetworkAddress` to conditionally return addresses no matter if a public uri is present in the call to LND.

Since we are making `EXTERNAL_ADDRESS` optional in another PR, we need to stop erroring if LND does not have a uri available. 

Additionally, i've removed the option for `includeHost` as its not used in any of our code

Related PR: https://github.com/sparkswap/lnd-engine/pull/181


dannypaz:

src/engine-actions/get-payment-channel-network-address.js
```diff
@@ -11,27 +11,24 @@ const HOST_DELIMITER = '@'
...
-    throw new Error(`LND has no uri avaialble`)
+async function getPaymentChannelNetworkAddress () {
+  const {
+    identityPubkey,
+    uris = []
+  } = await getInfo({ client: this.client })
+
+  // The uri will only exist if a user has set the `--external-ip` flag on the
+  // LND instance. If this uri does not exist, we will simply use the lnd public key
+  if (!uris.length) {
```
martineehrlich: Will `identityPubkey` always exist?

dannypaz: @martineehrlich yeah it should always exist.

dannypaz: @martineehrlich just to shield us from API changes, its probably worth it to check if it exists anyways

PR #188, opened 2019-04-04
Remove macaroon warning

treygriffith:
This change removes the macaroon warning if the engine is just starting up (unknown state) or does not have wallets created, in which case macaroons are not expected to exist.

It should make the initial startup less confusing for users.


src/lnd-setup/generate-lightning-client.js
```diff
@@ -17,14 +18,20 @@ const fs = require('fs')
...
-function generateLightningClient ({ host, protoPath, tlsCertPath, macaroonPath, logger }) {
+function generateLightningClient ({ host, protoPath, tlsCertPath, macaroonPath, status, logger }) {
   const { lnrpc } = loadProto(protoPath)
 
   const macaroonExists = fs.existsSync(macaroonPath)
 
+  // A macaroon file should exist if the Wallets have been created, so if the engine is
+  // in any state other than NEEDS_WALLET or UNKNOWN, we should show a warning.
+  // UNKNOWN is exempted since the engine could be in a NEEDS_WALLET state.
+  const showMacaroonWarn = ![ ENGINE_STATUSES.UNKNOWN, ENGINE_STATUSES.NEEDS_WALLET ].includes(status)
```
dannypaz: [nit] would define this closer to where it is used

PR #189, opened 2019-04-04
Feature - Add getPeers

dannypaz:
This PR adds a `getPeers` method for lnd-engine that exposes all active peers connected to the engine.


src/engine-actions/get-peers.spec.js
```diff
@@ -0,0 +1,41 @@
...
+const getPeers = rewire(path.resolve(__dirname, 'get-peers'))
+
+describe('get-peers', () => {
+  let listPeersStub
+  let peers
+  let client
+
+  beforeEach(() => {
+    peers = [
+      { pubKey: '1234', address: '192.168.0.1:10009', inbound: false },
```
treygriffith: do peers have an `inbound` property? What does it mean?

dannypaz: @treygriffith they do. I added the field here to make sure that we are formatting peers to only include pubKey and address for now

PR #191, opened 2019-04-04
stop logging host and instead use pubkey

dannypaz:
This PR makes changes to log messages in `createChannel` where we are logging the host, however the host may not be available at all times. Instead we should log the pubkey


martineehrlich:
did you miss https://github.com/sparkswap/lnd-engine/pull/191/files#diff-b7e454eb5526d0778b8abb6f047c155dR81

dannypaz:
@martineehrlich yess good catch

dannypaz:
[Prevent LND-Engine from logging the host](https://trello.com/c/ed7sNLIv/1390-prevent-lnd-engine-from-logging-the-host)

PR #197, opened 2019-04-10
Feature - Add total reserved channel balance

bradleystachurski:
This PR adds an engine action `getTotalReservedChannelBalance`, which returns the total balance of reserved funds across all channels where the daemon is an initiator. For LND, this balance represents the total `commitFees`, however this may include other balances for other payment channel implementations.

Related: https://github.com/sparkswap/broker/pull/486


src/engine-actions/get-commit-fees.js
```diff
@@ -0,0 +1,27 @@
...
+  console.log('channels', channels)
+
+  if (channels.length === 0) {
+    this.logger.debug('getCommitFees: No channels exist')
+    return Big(0).toString()
+  }
+
+  const totalCommitFees = channels.reduce((acc, c) => {
+    // We include commit fees only if the channel was the initiator
+    const commitFee = c.initiator ? Big(c.commitFee) : Big(0)
```
martineehrlich: I might do a filter on whether or not the channel is initiator and then not do this ternary.

bradleystachurski:
[Include commit fees in broker balances](https://trello.com/c/fOUJpcpJ/1333-include-commit-fees-in-broker-balances)

src/engine-actions/get-total-commit-fees.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+async function getTotalCommitFees () {
+  const { channels = [] } = await listChannels({ client: this.client })
+
+  if (channels.length === 0) {
+    this.logger.debug('getCommitFees: No channels exist')
+    return Big(0).toString()
+  }
+
+  // We include commit fees only if the channel was the initiator
```
treygriffith: Would be better to include an explanation of why that is rather than just what we're doing.

src/engine-actions/get-total-commit-fees.js
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Get total commit fees for all channels for a specific daemon
+ *
+ * @returns {string} totalCommitFees (int64)
+ */
+async function getTotalCommitFees () {
+  const { channels = [] } = await listChannels({ client: this.client })
+
+  if (channels.length === 0) {
```
treygriffith: This seems like an unnecessary step since you're doing a reduce - won't it just return `Big(0).toString()` even if there are no channels returned?

src/engine-actions/get-total-reserved-channel-balance.js
```diff
@@ -0,0 +1,27 @@
...
+  const { channels = [] } = await listChannels({ client: this.client })
+
+  // We filter for channels where the engine was the initiator because available balances for
+  // a channel are calculated less commit fees when a party was the initiator. The rationale of
+  // this calculation is to put the responsibility on the initiator to pay for a force-close
+  const initiatorChannels = channels.filter(c => c.initiator)
+  this.logger.debug(`getTotalReservedChannelBalance: ${channels.length} channels exist, ${initiatorChannels.length} channels as initiator`)
+
+  const totalReservedChannelBalance = initiatorChannels.reduce((acc, c) => {
+    return acc.plus(Big(c.commitFee))
```
treygriffith: you don't need to `Big(...)` inside of a `Big` function, it does that for you.

PR #198, opened 2019-04-10
Feature - getChainTransactions

dannypaz:
This PR adds a new api `getChainTransactions` that produces an array of on-chain transactions for the given node.

Types of Transactions:
1. Depositing to wallet
2. Withdrawl
3. Creating channels
4. Releasing channels


dannypaz:
[Record funds spent by the broker opening channels](https://trello.com/c/g86thBm8/1332-record-funds-spent-by-the-broker-opening-channels)

src/engine-actions/get-chain-transactions.js
```diff
@@ -0,0 +1,194 @@
...
+    // calculate the fee based on information from the closing transaction
+    if (type === TRANSACTION_TYPES.CHANNEL_CLOSE) {
+      fees = Big(closedChannelFees.get(txHash)).div(this.quantumsPerCommon)
+    }
+
+    return {
+      type,
+      amount: Big(amount).div(this.quantumsPerCommon).toString(),
+      transaction: txHash,
+      blockNumber: blockHeight,
```
treygriffith: why `blockNumber` instead of `blockHeight`?

src/engine-actions/get-chain-transactions.js
```diff
@@ -0,0 +1,194 @@
...
+      timestamp: timeStamp,
+      fees: fees.toString()
```
treygriffith: does this need to be `toFixed()`?

src/engine-actions/get-chain-transactions.js
```diff
@@ -0,0 +1,194 @@
...
+ * @constant
+ * @type {Object}
+ * @default
+ */
+const TRANSACTION_TYPES = Object.freeze({
+  CHANNEL_OPEN: 'CHANNEL_OPEN',
+  CHANNEL_CLOSE: 'CHANNEL_CLOSE',
+  DEPOSIT: 'DEPOSIT',
+  WITHDRAW: 'WITHDRAW',
+  PENDING: 'PENDING',
```
treygriffith: Is `PENDING` really a transaction type? Seems more like a status of a transaction which has some other type (e.g. a pending channel open transaction vs a confirmed channel open transaction vs a pending channel close transaction)

treygriffith:
Really like this PR 👏

src/engine-actions/get-chain-transactions.js
dannypaz: @treygriffith I can try to be more specific, but used PENDING just because I didn't want to add all of the specific types yet. PENDING is basically "waiting to be put into a block"

src/engine-actions/get-chain-transactions.js
dannypaz: @treygriffith no reason. Im cool changing to blockHeight

src/engine-actions/get-chain-transactions.js
dannypaz: Honestly, its probably worth it to be specific, let me see what we can do

src/engine-actions/get-chain-transactions.js
```diff
@@ -0,0 +1,190 @@
...
+})
+
+/**
+ * Gets all opening and closing transaction ids from a list of channels.
+ * @param {Object} channels
+ * @returns {Set} channelTxIds
+ */
+async function getTxIdsForChannels (channels) {
+  return channels.reduce((channelTxIds, { closingTxHash, channelPoint }) => {
+    const [ openingTxHash, _ ] = channelPoint.split(':') // eslint-disable-line
```
treygriffith: I don't think you need the second parameter here

src/lnd-actions/get-transactions.js
```diff
@@ -0,0 +1,21 @@
...
+ * @see http://api.lightning.community/#getTransactions
+ * @param {Object} opts
+ * @param {LndClient} opts.client
+ * @returns {Array<Object>} response
+ */
+function getTransactions ({ client }) {
+  return new Promise((resolve, reject) => {
+    client.getTransactions({}, { deadline: deadline() }, (err, res) => {
+      if (err) return reject(err)
+      const { transactions = [] } = res
```
treygriffith: do we do these kinds of transformations in other lnd actions? seems kind of risky since this code is largely untested

src/engine-actions/get-chain-transactions.js
```diff
@@ -0,0 +1,190 @@
...
+    // that the transaction has not yet been placed into a block (no confirmations)
+    const isPending = (Big(blockHeight).eq(0) || pendingChannelTxIds.has(txHash))
+
+    // `totalFees` will be zero if this is a CHANNEL_CLOSE transaction type
+    let fees = Big(totalFees)
+
+    // If we've identified the type of transaction as channel close then `totalFees`
+    // will be 0. We calculate the fee based on information from the closing transaction
+    // but this is only available if the transaction has hit the blockchain
+    if (type === TRANSACTION_TYPES.CHANNEL_CLOSE && !isPending) {
```
treygriffith: don't you need to check if we were the initiator?

dannypaz: @treygriffith We do need to check for the initiator, however we don't currently have the information available from LND. Ive opened a PR to add initiator to the closed channel summary, however for now this should be OK

https://github.com/sparkswap/lnd/pull/34

dannypaz: Going to add a TODO to this code

PR #209, opened 2019-05-06
Feature - Add `changeWalletPassword` method to lnd-engine

dannypaz:
This adds a public method to lnd engine `changeWalletPassword` that allows consumers to change the wallet passwords of an engine


dannypaz:
[Change Password RPC for broker](https://trello.com/c/FC87novu/1542-change-password-rpc-for-broker)

PR #213, opened 2019-06-04
Get local channel balance with a given peer

treygriffith:
This change adds `getTotalBalanceForAddress`, a higher level method for retrieving the total local balance in all pending and open channels with a peer specified by a Payment Channel Network address.


src/engine-actions/get-total-balance-for-address.js
```diff
@@ -0,0 +1,24 @@
...
+ * Get local balance of all channels for a specific address
+ * @param {string} address
+ * @returns {string} totalBalance (int64)
+ */
+async function getTotalBalanceForAddress (address) {
+  const channelsForAddress = await getChannelsForRemoteAddress.call(this, address)
+
+  if (channelsForAddress.length === 0) {
+    this.logger.debug('getTotalBalanceForAddress: No open or pending channels exist', { symbol: this.symbol, address })
+    return Big(0).toString()
```
bradleystachurski: [nit] Why use `Big(0).toString()` instead of just '0'?

src/engine-actions/get-total-balance-for-address.spec.js
```diff
@@ -0,0 +1,58 @@
...
+  })
+
+  afterEach(() => {
+    revertChannels()
+  })
+
+  it('gets all channels for the given address', async () => {
+    await getTotalBalanceForAddress(address)
+    expect(getChannelsForRemoteAddressStub).to.have.been.called()
+    expect(getChannelsForRemoteAddressStub).to.have.been.calledWith(address)
```
bradleystachurski: [nit] Do we also want to test the context of `this` for the call?

src/engine-actions/get-total-balance-for-address.js
treygriffith: More consistent with other values, but it is really no different.

PR #215, opened 2019-06-05
Add minimum channel size

treygriffith:
This change adds the notion of a minimum channel size to both LND Engine and LND itself.

In LND, this configuration is used to reject incoming channels that are too small to be useful because of fees.

In the LND Engine, this configuration is used to prevent opening channels that are too small to be useful.


dannypaz:
[Add minimum channel balance to lnd-engine](https://trello.com/c/4dBuHnmo/281-add-minimum-channel-balance-to-lnd-engine)

docker/lnd/conf/lnd-bitcoind.conf
```diff
@@ -14,6 +14,7 @@ debuglevel=info
 maxpendingchannels=2
 nobootstrap=1
 tlsextradomain=lnd_btc
+minchansize=250000
```
dannypaz: @treygriffith I was having trouble thinking of how often we would want these values to change. 

We could make this an ENV instead of throwing it into the `.conf` file so that we can edit on the fly. If that happened, then we would remove `min` and `max` channel balance from the configuration and just let LND handle the min/max which might be better for us in the long run, so it can be a dynamic number and fluctuate with the market.

We can punt this part for now, but wanted to get your thoughts

dannypaz:
@treygriffith forgot to ask. This code would affect how we are opening channels for a larger balance (say 1 BTC). There is a potential to reject a channel being created from a user. Do we need to expose `min` and `max` so that the broker can handle this case?

treygriffith:
@dannypaz yes the idea is to expose this to the Broker to use.

PR #217, opened 2019-06-05
Open multiple channels for a single funding amount

treygriffith:
This change adds a new engine action to open multiple channels to satisfy a given funding amount rather than opening a single channel.

### Todos
- [x] Tests
- [x] Remove `createChannel`
- [x] Docs


src/engine-actions/create-channels.js
```diff
@@ -0,0 +1,186 @@
...
+    maxChannelBalance,
+    minChannelBalance,
+    quantumsPerCommon,
+    symbol,
+    logger
+  } = engine
+
+  let amount = Big(fundingAmount)
+
+  // our last channel will be our only channel not at maxChannelBalance, so we need to handle it separately
```
bradleystachurski: Nice comment, this is much easier to understand how the next chunk of logic should work without context.

src/engine-actions/create-channels.js
```diff
@@ -0,0 +1,186 @@
...
+    maxChannelBalance,
+    quantumsPerCommon,
+    feeEstimate,
+    symbol,
+    logger
+  } = engine
+
+  const amount = Big(fundingAmount)
+
+  // round up to find the number of channels to open
```
bradleystachurski: Is there a reason you dropped the link to the `round` docs? Might be useful for folks looking at the code for the first time.

That said, it's also obvious this is a method on `big.js`.

src/engine-actions/create-channels.js
```diff
@@ -0,0 +1,186 @@
...
+  const { publicKey, host } = networkAddressFormatter.parse(paymentChannelNetworkAddress)
+  const loggablePublicKey = loggablePubKey(publicKey)
+
+  logger.debug(`Attempting to create ${channelsToOpen} channels with ${loggablePublicKey}`)
+
+  await connectPeer(publicKey, host, { client, logger })
+
+  logger.debug(`Successfully connected to peer: ${loggablePublicKey}`)
+
+  for (var i = 0; i < channelsToOpen; i++) {
```
bradleystachurski: I overheard part of a conversation that we use a `for` loop here instead of a `Promise.all` since the behavior could get weird trying to open multiple channels concurrently.

If that is the reasoning for using the `for` loop, it might be worth adding a comment.

src/constants/channel-rounding.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * List of rounding behaviors when opening multiple channels and encountering a channel that
+ * would be uneconomically small.
+ *
+ * @constant
+ * @type {Object<key, String>}
```
dannypaz: lowercase `string`. unfortunately jsdoc doesnt do a good job for complex types like this

src/engine-actions/create-channels.js
```diff
@@ -0,0 +1,186 @@
...
+const getUncommittedBalance = require('./get-uncommitted-balance')
+const { CHANNEL_ROUNDING } = require('../constants')
+
+/**
+ * Determine the amount to fund channels with by taking into account
+ * uneconomic channels (that are too small to be worth it.)
+ * @param   {Engine} engine
+ * @param   {string} fundingAmount - int64 string of amount desired to open
+ * @param   {number} roundBehavior - Behavior to use when encoutnering a channel that is too small.
+ * @returns {Big}                    Modified amount to open in channels
```
bradleystachurski: I'd assume we're okay returning `Big` since this is a private method?

src/engine-actions/create-channels.js
dannypaz: [nit] I _think_ you can use `let` here

src/engine-actions/create-channels.js
```diff
@@ -0,0 +1,186 @@
...
+      `Maximum estimated funding amount: ${maxFundingAmount} ${symbol}.`)
+  }
+}
+
+/**
+ * Executes a connection and opens channels w/ another lnd instance
+ *
+ * @param {string} paymentChannelNetworkAddress
+ * @param {string} fundingAmount - int64 string
+ * @param {number} roundBehavior - Behavior when encountering a channel that is too small
```
dannypaz: string here

src/engine-actions/create-channels.js
treygriffith: Yes that's the idea.

PR #218, opened 2019-06-06
Upgrade to non-fork LND

dannypaz:
In order for lnd-engine to upgrade to the official LND release, we need to modify logic regarding invoices and swaps to handle the new API.

This change IS NOT backwards compatible. If YOU update your lnd versions with this PR, you WILL NOT be able to revert back to a previous engine version

**Changes**
1. Updates Dockefile for lnd
2. updates `build.sh` for new lnd
3. updates `status` checking for engine


treygriffith:
It seems like these changes (e.g. removing the extpreimage arg) are not backward compatible. Is that the case? If so, why have it be a build flag at all? (shouldn't it just be in a branch or similar?)

dannypaz:
[Upgrade LND to 0.6.1](https://trello.com/c/xf2I5NPl/1632-upgrade-lnd-to-061)

treygriffith:
This looks good to me and as a staging branch for future changes.

dannypaz:
Merging changes into master LND as we prepare to upgrade broker w/ lnd

PR #222, opened 2019-06-13
Add outbound option to getTotalBalanceForAddress

clark800:
The outbound option determines whether localBalance or remoteBalance is used in the calculation of the total balance.


package.json
```diff
@@ -71,5 +71,10 @@
     "@grpc/proto-loader": "0.5.0",
     "big.js": "5.2.2",
     "grpc": "1.19.0"
+  },
+  "standard": {
+    "env": [
+      "mocha"
```
dannypaz: @clark800 was your editor giving you issues? We might not have updated the docs saying we dont use standard anymore. The plugin would be eslint

clark800: Ok, is there a custom eslintrc file or are you just using the eslint default configuration?

clark800: @dannypaz updated

PR #223, opened 2019-06-17
Feature - Recover wallet during wallet creation

dannypaz:
Builds off of <https://github.com/sparkswap/lnd-engine/pull/218>

This PR adds functionality to the lnd-engine to allow users to recover a wallet through the `createWallet` command.


dannypaz:
[add recoverWallet method to lnd-engine](https://trello.com/c/31DVVnE8/982-add-recoverwallet-method-to-lnd-engine)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "lnd-engine",
-  "version": "0.5.4-beta-rc2",
+  "version": "0.5.5-beta-rc1",
```
treygriffith: Do you want to do the bump here or in a separate PR? I thought in general we were doing these in separate PRs.

src/engine-actions/create-wallet.js
```diff
@@ -3,25 +3,57 @@ const {
...
+ * @param {Buffer} opts.backup - binary string of an lnd backup file
  * @returns {Array<string>} 24 word cipher seed mnemonic
  */
-async function createWallet (password) {
-  const { cipherSeedMnemonic } = await genSeed({ client: this.walletUnlocker })
-
-  if (typeof password !== 'string') {
-    throw new Error('Provided password must be a string value')
+async function createWallet (password, { seed, backup, recoveryWindow = undefined } = {}) {
+  if (!seed && backup) {
```
treygriffith: What happens if they provide the seed but not the backup?

src/engine-actions/create-wallet.js
```diff
@@ -3,25 +3,57 @@ const {
...
+  return true
+}
+
 /**
  * Creates a wallet
  *
  * @param {string} password - wallet password, used to unlock lnd wallet
+ * @param {Object} opts
+ * @param {Array<string>} opts.seed - 24 word cipher seed mnemonic
+ * @param {Buffer} opts.backup - binary string of an lnd backup file
```
treygriffith: Is this a buffer or a string? Your comment is ambiguous

src/engine-actions/create-wallet.js
```diff
@@ -3,25 +3,57 @@ const {
   initWallet
 } = require('../lnd-actions')
 
+/**
+ * Wallet recovery window number in blocks
+ * @constant
+ * @type {number}
+ * @default
+ */
+const RECOVERY_WINDOW_DEFAULT = 5000
```
treygriffith: Why this number?

src/engine-actions/create-wallet.js
```diff
@@ -3,25 +3,57 @@ const {
...
+ * @param {Array<string>} opts.seed - 24 word cipher seed mnemonic
+ * @param {Buffer} opts.backup - binary string of an lnd backup file
  * @returns {Array<string>} 24 word cipher seed mnemonic
  */
-async function createWallet (password) {
-  const { cipherSeedMnemonic } = await genSeed({ client: this.walletUnlocker })
-
-  if (typeof password !== 'string') {
-    throw new Error('Provided password must be a string value')
+async function createWallet (password, { seed, backup, recoveryWindow = undefined } = {}) {
```
treygriffith: Did you consider making this two different engine actions? Although they are the same (`initWallet`) on LND, they feel pretty different to me from a user (and implementation) perspective.

src/engine-actions/create-wallet.js
dannypaz: will only restore on-chain funds, which is OK. but you _need_ the seed for a backup file

src/engine-actions/create-wallet.js
dannypaz: it's a buffer (the file should already be open and sent). I didnt want to use `fs` in this function

src/engine-actions/create-wallet.js
dannypaz: @treygriffith I did have it as a second action. TBH it would be better that way. will make the change

package.json
dannypaz: @treygriffith we should bump the version now because you'll have conflicts when working w/ older images of the broker vs. newer images of the relayer. It makes it easier for the time being, while the broker is on an older lnd version

src/engine-actions/create-wallet.js
dannypaz: Arbitrary default. This is a wallet address look-ahead. This recovery window will go back 5000 addresses that were generated from the seed. It seemed like a decent number for default until someone has an issue.

package.json
dannypaz: sorry, i misread. I can bump the version in a different PR if you wanted

src/engine-actions/recover-wallet.js
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @param {string} password - wallet password, used to unlock lnd wallet
+ * @param {Array<string>} seed - 24 word cipher seed mnemonic
+ * @param {Buffer} backup - Buffer of the LND static channel backup file
+ * @param {number} recoveryWindow - number in blocks for look back
+ * @returns {Array<string>} 24 word cipher seed mnemonic
+ */
+async function recoverWallet (password, seed, backup, recoveryWindow = undefined) {
+  // The seed is required if the user is using a backup file
+  if (!seed && backup) {
```
treygriffith: I might add a warning if they are using only the seed to let them know that only on-chain funds will be recovered.

src/engine-actions/recover-wallet.js
```diff
@@ -0,0 +1,56 @@
...
+  // Password must be converted to buffer in order for lnd to accept
+  // as it does not accept String password types at this time.
+  const walletPassword = Buffer.from(password, 'utf8')
+
+  // If the user has provided a seed, then we will try to recover the wallet
+  // on creation
+  await isValidSeed(seed)
+
+  // If a recovery window was not provided, then we will use a default value
+  recoveryWindow = recoveryWindow || RECOVERY_WINDOW_DEFAULT
```
treygriffith: [nit] Debugger on this value would be nice

PR #224, opened 2019-06-17
FEATURE - Add dynamic backup directory configuration

dannypaz:
Builds off of <https://github.com/sparkswap/lnd-engine/pull/218>

This PR adds a `backupfilepath` configuration setting to lnd that specifies `/backup/lnd_<symbol>.backup` as the static channel backup file for the container.

Additionally, with the backup and the wallet seed, we will (almost) be able to entirely restore the wallet for the broker. The caveat is this change does not backup payments that are in-flight (HTLC)


dannypaz:
[add recoverWallet method to lnd-engine](https://trello.com/c/31DVVnE8/982-add-recoverwallet-method-to-lnd-engine)

docker/lnd/start-lnd-btc.sh
```diff
@@ -33,4 +33,9 @@ if [[ -n "$LND_FEE_RATE" ]]; then
     PARAMS="$PARAMS --bitcoin.feerate=$LND_FEE_RATE"
 fi
 
+if [[ -n "$BACKUP_DIR" ]]; then
```
treygriffith: Are there cases where we won't specify this?

dannypaz: I cant think of any when we (sparkswap) wouldn't but if it is left out, itll just default to the `data` dir.

treygriffith: It might pay to be a bit more opinionated here and not support not customizing this. Would avoid some weird issues where this value is not defined by mistake.

PR #226, opened 2019-06-18
Allows local Linux setup

zallarak:
Using `host.docker.internal` to access the host machine does not work on Linux. This naively fixes it. Code changes must also be made to `sparkswap/broker`.


dannypaz:
@zallarak was this tested on mac/windows? If not, we can do this in the morning to make sure the changes are A-OK (they should be but we should test JIC)

docker/bitcoind/start-bitcoind.sh
```diff
@@ -85,6 +83,11 @@ fi
 # an IP and then connect to that machine
 if [[ "$NETWORK" == "regtest" ]] && [[ -n "$CONNECT_HOST" ]]; then
     if [[ "$CONNECT_HOST" == "host.docker.internal" ]]; then
+        ping -c1 $CONNECT_HOST > /dev/null 2>&1
```
dannypaz: I would make a note in each place describing why we are adding an extra ip/host to our host file on all these boxes. Would emphasize that this is only used because we do networking with `host.docker.internal` which is not currently supported for linux boxes

zallarak: Will do. Also, if you think of a clean way to do this that doesn't duplicate so much code, let me know (we do the same thing in 4 places in this repo and 1 place in `lnd-engine` ...) let me know. I haven't thought of a clean way so far.

zallarak:
Comments addressed. Pending: testing on Mac.

dannypaz:
[Linux Local Regtest Setup](https://trello.com/c/DjKFyt2R/1705-linux-local-regtest-setup)

docker/bitcoind/start-bitcoind.sh
```diff
@@ -85,6 +83,14 @@ fi
 # an IP and then connect to that machine
 if [[ "$NETWORK" == "regtest" ]] && [[ -n "$CONNECT_HOST" ]]; then
     if [[ "$CONNECT_HOST" == "host.docker.internal" ]]; then
+        # Using host.docker.internal to access the host IP does not work for Linux
+        # This is a known issue with Docker: https://github.com/docker/for-linux/issues/264
+        # Here, we manually map host.docker.internal to the host IP in /etc/hosts
```
dannypaz: same comment as here: https://github.com/sparkswap/broker/pull/544/files#r295015291

PR #228, opened 2019-06-20
Use Hold Invoice in prepareSwap

clark800:
Use LND's Hold Invoice RPC in prepareSwap


src/lnd-setup/generate-lightning-client.js
```diff
@@ -23,7 +23,7 @@ const { ENGINE_STATUSES } = require('../constants')
  * @returns {grpc.Client} lnrpc Lightning client definition
  */
 function generateLightningClient ({ host, protoPath, tlsCertPath, macaroonPath, status, logger }) {
-  const { lnrpc } = loadProto(protoPath)
+  const { lnrpc, invoicesrpc } = loadProto(protoPath, ['rpc.proto', 'invoicesrpc/invoices.proto'])
```
treygriffith: is there any way we can make these path strings variables? Either as constants or passed in from the caller? Feels a little weird as is.

src/engine-actions/prepare-swap.js
```diff
@@ -34,19 +34,12 @@ async function prepareSwap (orderId, swapHash, value) {
...
+  const { invoice } = await addHoldInvoice(params, { client: this.client })
 
-  if (rHash !== swapHash) {
-    throw new Error(
-      `Error while preparing for swap: returned rHash did not match passed swapHash: rHash=${rHash}, swapHash=${swapHash}`
-    )
-  }
-
-  return swapHash
+  return invoice
```
treygriffith: Why change the return value here? Why not keep it as the hash?

clark800: My thinking was that the caller already has `swapHash` since it was passed in, and the invoice is what is being generated in this method and will be lost if we don't return it (or would have to be looked up some other way). Since the call site isn't using the return value it looks like we can change this without affecting anything else. I need to fix this either way though because it's supposed to be called `paymentRequest`.

src/lnd-setup/generate-lightning-client.js
clark800: Updated

src/engine-actions/prepare-swap.js
treygriffith: Ok that makes sense.

dannypaz:
@clark800 description is needed on this pr

src/lnd-setup/generate-lightning-client.js
```diff
@@ -60,7 +65,9 @@ function generateLightningClient ({ host, protoPath, tlsCertPath, macaroonPath,
     rpcCredentials = grpc.credentials.combineChannelCredentials(sslCredentials, macaroonCredentials)
   }
 
-  return new lnrpc.Lightning(host, rpcCredentials, {})
+  const client = new lnrpc.Lightning(host, rpcCredentials, {})
+  client.invoices = new invoicesrpc.Invoices(host, rpcCredentials, {})
```
treygriffith: probably worth testing the exposure of this new client

src/lnd-setup/generate-lightning-client.js
```diff
@@ -60,7 +65,9 @@ function generateLightningClient ({ host, protoPath, tlsCertPath, macaroonPath,
     rpcCredentials = grpc.credentials.combineChannelCredentials(sslCredentials, macaroonCredentials)
   }
 
-  return new lnrpc.Lightning(host, rpcCredentials, {})
+  const client = new lnrpc.Lightning(host, rpcCredentials, {})
```
treygriffith: is this kept as the base client to prevent too many cascading changes in other parts of this and other repos?

src/lnd-setup/generate-wallet-unlocker-client.js
```diff
@@ -18,7 +20,7 @@ const fs = require('fs')
  * @returns {Object} lnrpc WalletUnlocker client definition
  */
 function generateWalletUnlockerClient ({ host, protoPath, tlsCertPath }) {
-  const { lnrpc } = loadProto(protoPath)
+  const { lnrpc } = loadProto(protoPath, PROTO_FILES)
```
dannypaz: might need to update spec to make sure we are passing in PROTO_FILES

src/lnd-setup/generate-wallet-unlocker-client.js
```diff
@@ -7,6 +7,8 @@ const grpc = require('grpc')
 const loadProto = require('../utils/load-proto')
 const fs = require('fs')
 
+const PROTO_FILES = ['rpc.proto']
```
dannypaz: freeze it and add jsdoc

proto/rpc.proto
```diff
@@ -1,6 +1,6 @@
 syntax = "proto3";
 
-
+import "google/api/annotations.proto";
```
dannypaz: For historical purposes, we actually remove this line w/ `sed` in the build.sh. but since we changed the name of this file (from lnd-rpc to rpc), this change is now added.

We should remove the associated change in build.sh here: https://github.com/sparkswap/lnd-engine/blob/master/scripts/build.sh#L42-L44

src/lnd-setup/generate-lightning-client.js
```diff
@@ -9,6 +9,11 @@ const loadProto = require('../utils/load-proto')
 const fs = require('fs')
 const { ENGINE_STATUSES } = require('../constants')
 
+const PROTO_FILES = [
```
dannypaz: `Object.freeze` this and add jsdoc

src/lnd-setup/generate-lightning-client.js
clark800: Yeah, it was causing trouble the other way and I figured that could be a separate refactoring if it's considered important enough.

clark800:
@treygriffith @dannypaz updated

PR #229, opened 2019-06-24
Add expiry parameters to prepareSwap and make it idempotent

clark800:
Change `prepareSwap` to be ready for use of hold invoices in swaps by adding the `expiry` and `cltvExpiry` parameters and making it idempotent.


src/engine-actions/prepare-swap.js
```diff
@@ -23,23 +14,32 @@ const MEMO_PREFIX = 'sparkswap-swap-pivot:'
...
- * @returns {string} swapHash      Hash that will be translated
+ * @param {string} swapHash - swap hash that will be associated with the swap
+ * @param {string} value - Int64 string of the value of inbound currency
+ * @param {number} expiry - payment request expiry time in seconds
+ * @param {number} cltvExpiry - delta to use for the time-lock of the CLTV
+ *                              extended to the final hop (in blocks)
+ * @returns {string} paymentRequest - lightning invoice for a payment
  */
-async function prepareSwap (orderId, swapHash, value) {
+async function prepareSwap (orderId, swapHash, value, expiry, cltvExpiry) {
```
treygriffith: In https://github.com/sparkswap/broker/pull/547 we're no longer passing `orderId` as a parameter.

src/engine-actions/prepare-swap.js
```diff
@@ -23,23 +14,32 @@ const MEMO_PREFIX = 'sparkswap-swap-pivot:'
...
-    memo: `${MEMO_PREFIX}${orderId}`,
-    value,
-    expiry: SWAP_EXPIRY,
-    hash: swapHash
-  }
-
-  const { paymentRequest } = await addHoldInvoice(params, { client: this.client })
+  // make prepareSwap idempotent by returning the existing invoice if one exists
+  try {
+    const { paymentRequest } = await lookupInvoice(swapHash, { client: this.client })
```
dannypaz: does the call to lookupInvoice throw an error or is it the destructuring? I ask because we will create an invoice _for any_ error that occurs, which might have consequences since we only want to add an invoice if it's not found.

clark800:
[Use absolute and relative expiration in `prepareSwap`, and make it idempotent](https://trello.com/c/WahnbYVy/1730-use-absolute-and-relative-expiration-in-prepareswap-and-make-it-idempotent)

src/engine-actions/prepare-swap.js
clark800: It looks like `lookupInvoice` throws an error when the invoice is not found (the documentation isn't really clear but I looked at the source). I don't think the destructuring can produce an error here. There could be an error if `swapHash` is invalid, but that should fail in the catch block also.

src/engine-actions/prepare-swap.js
clark800: So we don't need it in the memo anymore?

clark800:
@treygriffith updated

src/engine-actions/prepare-swap.js
```diff
@@ -17,29 +8,38 @@ const SWAP_EXPIRY = '3600'
...
+ *                              extended to the final hop (in seconds)
+ * @returns {string} paymentRequest - lightning invoice for a payment
  */
-async function prepareSwap (orderId, swapHash, value) {
+async function prepareSwap (swapHash, value, expiryTime, cltvExpiry) {
   this.logger.info(`Preparing swap for ${swapHash}`, { value })
+  // make prepareSwap idempotent by returning the existing invoice if one exists
+  try {
+    const { paymentRequest } = await lookupInvoice(swapHash, { client: this.client })
+    return paymentRequest
```
treygriffith: shouldn't we validate the parameters of the invoice here before returning? Seems like there is a risk that we think we created an invoice with a set of parameters, but the already existing invoice had different parameters.

In particular, making sure this is a hold invoice and has the correct amount and cltv expiry.

(This gets complicated with the expiry time - I haven't thought through that).

src/engine-actions/prepare-swap.js
```diff
@@ -17,29 +8,38 @@ const SWAP_EXPIRY = '3600'
...
+  } catch (e) {
+    const expiry = Math.ceil((expiryTime - new Date()) / 1000)
+    const cltvExpiryBlocks = Math.ceil(cltvExpiry / this.secondsPerBlock)
```
treygriffith: [nit] A comment describing why we round up blocks here would be helpful.

src/engine-actions/prepare-swap.js
```diff
@@ -17,29 +8,38 @@ const SWAP_EXPIRY = '3600'
...
  */
-async function prepareSwap (orderId, swapHash, value) {
+async function prepareSwap (swapHash, value, expiryTime, cltvExpiry) {
   this.logger.info(`Preparing swap for ${swapHash}`, { value })
+  // make prepareSwap idempotent by returning the existing invoice if one exists
+  try {
+    const { paymentRequest } = await lookupInvoice(swapHash, { client: this.client })
+    return paymentRequest
+  } catch (e) {
+    const expiry = Math.ceil((expiryTime - new Date()) / 1000)
```
treygriffith: Seems like we would want to round this down, right? We want it to expire no later than the provided date?

src/engine-actions/prepare-swap.js
treygriffith: It was always a nice to have for debugging, but I think the hash is good enough.

src/engine-actions/prepare-swap.js
dannypaz: It looks like lookupinvoice will send a code `2` with the message `Error: 2 UNKNOWN: there are no existing invoices`. We should match on this error and only add a new invoice if it's this code.

clark800: Is that documented?

src/engine-actions/prepare-swap.js
clark800: Ok, I guess it doesn't hurt to be more defensive. Is there a way to check if an invoice is a hold invoice?

clark800:
@treygriffith updated

src/engine-actions/prepare-swap.js
dannypaz: Not documented. This is the functionality when you use a `swapHash` that doesn't exist. We do a similar check here: https://github.com/sparkswap/lnd-engine/blob/77b02eb69327477f123d49b71da865bdfce62809/src/engine-actions/get-status.js#L76

It is semi-fragile. GRPC has its own status codes, the only way to differentiate between the errors is through the message.

src/engine-actions/prepare-swap.js
```diff
@@ -17,29 +8,53 @@ const SWAP_EXPIRY = '3600'
...
+      // if the invoice doesn't match, we log an error and  prepare a new invoice
+      const hash = swapHash.toHex()
+      const message = `Invoice found for hash ${hash} but parameters are wrong`
+      this.logger.error(message)
+      throw new Error(message)
+    }
+    return paymentRequest
+  } catch (e) {
+    const expiry = Math.floor((expiryTime - new Date()) / 1000)
+    const cltvExpiryBlocks = Math.ceil(cltvExpiry / this.secondsPerBlock)
```
treygriffith: this looks like a duplicate definition to https://github.com/sparkswap/lnd-engine/pull/229/files#diff-23d78b84017667008c15e17b28942039R27

src/engine-actions/prepare-swap.js
```diff
@@ -17,29 +8,53 @@ const SWAP_EXPIRY = '3600'
...
-  const { paymentRequest } = await addHoldInvoice(params, { client: this.client })
+    if (invoiceValue !== value || invoiceCLTVExpiry !== cltvExpiryBlocks) {
+      // if the invoice doesn't match, we log an error and  prepare a new invoice
+      const hash = swapHash.toHex()
+      const message = `Invoice found for hash ${hash} but parameters are wrong`
+      this.logger.error(message)
+      throw new Error(message)
+    }
+    return paymentRequest
+  } catch (e) {
```
treygriffith: Why do this in every error case? Aren't there some errors where we would *not* want to create a new invoice?

src/engine-actions/prepare-swap.js
```diff
@@ -17,29 +8,53 @@ const SWAP_EXPIRY = '3600'
...
-    memo: `${MEMO_PREFIX}${orderId}`,
-    value,
-    expiry: SWAP_EXPIRY,
-    hash: swapHash
-  }
-
-  const { paymentRequest } = await addHoldInvoice(params, { client: this.client })
+    if (invoiceValue !== value || invoiceCLTVExpiry !== cltvExpiryBlocks) {
+      // if the invoice doesn't match, we log an error and  prepare a new invoice
+      const hash = swapHash.toHex()
```
treygriffith: I haven't seen this `toHex` method of a string. And do we only use it for logging? If so, sticking with base64 might be better.

src/engine-actions/prepare-swap.js
clark800: Discussed in person: since there's no documentation for error types, we can't safely distinguish error types, and it isn't likely to make much difference anyway.

PR #231, opened 2019-06-25
update translate swap to use max cltv from new SendPayment api

treygriffith:
This change updates `translateSwap` to use the new `cltv_limit` in `SendPayment` to find a route with a maximum timelock with an absolute time limit rather than querying the graph directly, which we did previously.

It's a requirement for https://github.com/sparkswap/broker/pull/547


src/engine-actions/translate-swap.js
```diff
@@ -1,220 +1,59 @@
...
 
 /**
  * Translates a swap to a new payment channel network by making a payment
  * to the counterparty (Taker) node on this network
  *
- * @param {string} takerAddress - Payment channel network address fo the Taker node we are paying
- * @param {string} swapHash     - swap hash that will be associated with the swap
- * @param {string} amount       - Int64 string of the amount of outbound currency in integer units
- * @param {string} extendedTimeLockDelta  - Int64 string of the time lock extended to us by on the first chain in seconds
+ * @param {string} takerAddress          - Payment channel network address of
```
dannypaz: @treygriffith is this a tool/formatting change for jsdoc stuff or are you doing this to break the comments up? I'd be curious to know if this actually parses correctly

src/lnd-actions/send-payment.js
```diff
@@ -14,17 +14,43 @@ const SEND_PAYMENT_DEADLINE = 30
...
+ *                                                 the route timelock
  * @param {Object} opts
  * @param {LndClient} opts.client - LND client to use
  * @returns {Promise<Object>} Resolves with the response from LND
  */
-function sendPayment ({ paymentRequest, paymentHash, destString, amt, finalCltvDelta, feeLimit }, { client }) {
-  const request = { paymentRequest, paymentHash, destString, amt, finalCltvDelta, feeLimit }
+function sendPayment (paymentOptions, { client }) {
+  const {
+    paymentRequest,
```
dannypaz: Can we make a note that `paymentRequest` is optional?

dannypaz:
@treygriffith love the cleanup in this PR. I've made some comments but they are minor. LGTM

src/engine-actions/translate-swap.js
treygriffith: I was trying to keep my lines shorter (80 characters). How can I check the parsing?

dannypaz: we can worry about when the time comes. no worries

src/engine-actions/translate-swap.js
```diff
@@ -5,6 +5,13 @@ const {
   DEFAULT_MIN_FINAL_DELTA
 } = CLTV_DELTA
 
+/**
+ * Default fee limit for swap routes.
+ * We expect to route swaps through a fee-less hub.
+ * @todo Make this value dynamic.
+ */
+const DEFAULT_FEE_LIMIT = 0
```
dannypaz: @treygriffith do we want to throw this in `config.json`?

src/engine-actions/translate-swap.js
```diff
@@ -1,220 +1,74 @@
...
-  const queryRoutesReq = {
-    pubKey: publicKey,
+  // Convert the maximum time into a time from now so we can convert it into
+  // blocks of relative time. We round down to a whole number of seconds to
+  // be conservative.
+  const timeFromNow = Math.floor((maxTime - new Date()) / 1000)
+
+  // Translate seconds into blocks, rounding down to ensure we can stay atomic.
+  // The LND api expects this to be an int32, which in javascript is a number.
+  const cltvLimit = parseInt(Big(timeFromNow).div(this.secondsPerBlock).round(3), 10)
```
clark800: `Big.round` returns a `Big` and `parseInt` expects a string, so this would be a type error; I think you could do
`parseInt(Big(timeFromNow).div(this.secondsPerBlock).round(0, 0).toFixed(0), 10)`

src/lnd-actions/send-payment.js
```diff
@@ -12,19 +12,45 @@ const SEND_PAYMENT_DEADLINE = 30
...
+    paymentHash,
+    destString,
+    amt,
+    finalCltvDelta,
+    feeLimit,
+    cltvLimit
+  } = paymentOptions
+
+  const request = {
```
clark800: What's the reason for unpacking and repacking here, are there potentially extra parameters being passed in?

src/engine-actions/translate-swap.js
```diff
@@ -1,220 +1,74 @@
...
-}
+const DEFAULT_FEE_LIMIT = 0
 
 /**
  * @typedef {Object} TranslateSwapOutcome
- * @property {String} paymentPreimage Base64 string of the preimage for the swapHash
- * @property {String} permanentError Error encountered that is permanent, and safe to cancel the upstream HTLC
+ * @property {String} paymentPreimage Base64 string of the preimage for the
+ *                                    swapHash
+ * @property {String} permanentError Error encountered that is permanent, and
```
clark800: Would it be possible to change this to an exception and just return the preimage? Using an object property to indicate errors is a little unconventional and makes it easier to make silent bugs when the caller forgets to check for the property.

src/engine-actions/translate-swap.js
treygriffith: I don't think so, because I expect that when we make this dynamic it will be by adding a parameter to the call rather than configuring it at startup. But I'm open to pushback.

src/lnd-actions/send-payment.js
treygriffith: None other than documentation, but I'll leave that in the jsdoc definition.

src/engine-actions/translate-swap.js
treygriffith: Good catch, thanks.

src/engine-actions/translate-swap.js
treygriffith: Yes, we could potentially have a specific error type, like we do in https://github.com/sparkswap/lnd-engine/pull/234.

treygriffith: One issue we may have with these custom error types is how they are used when we have multiple engines. Each engine will have a different custom error type, so type comparison won't work.

What do you think @clark800 ?

treygriffith: One idea would be to expose them on the instance of the engine, but that seems pretty nasty.

src/engine-actions/translate-swap.js
```diff
@@ -1,227 +1,80 @@
...
+
+  // Translate seconds into blocks, rounding up so that we're sure the end
+  // node will accept the payment. The LND api expects this to be an int32,
+  // so we convert to a javascript number.
+  const finalCltvDelta = parseInt(Big(finalCltvDeltaSecs).div(this.secondsPerBlock).round(0).toFixed(0), 10)
+
+  const {
+    paymentPreimage,
+    paymentError
+  } = await sendPayment({
```
bradleystachurski: Since `sendPayment` uses `SendPaymentSync` on lnd, our original conversation about calling `TrackPayment` first followed by `sendPayment` should be correct.

Alternatively, we could change `sendPayment` from calling `SendPaymentSync` on `rpc.proto` to calling `SendPayment` on `router.proto`, which errors if the payment is already in flight or has already been paid (https://github.com/lightningnetwork/lnd/blob/f03533c67a2b2dc62c81b7c7972316d5a98e32eb/lnrpc/routerrpc/router_server.go#L206-L207).

I don't think we gain much by changing to `SendPayment` on `router.proto`, however I'm open to making those changes when we add `TrackPayment`.

src/engine-actions/translate-swap.js
```diff
@@ -1,227 +1,80 @@
...
-  const queryRoutesReq = {
-    pubKey: publicKey,
+  // Convert the maximum time into a time from now so we can convert it into
+  // blocks of relative time. We round down to a whole number of seconds to
+  // be conservative.
+  const timeFromNow = Math.floor((maxTime - new Date()) / 1000)
+
+  // Translate seconds into blocks, rounding down to ensure we can stay atomic.
+  // The LND api expects this to be an int32, which in javascript is a number.
+  const cltvLimit = parseInt(Big(timeFromNow).div(this.secondsPerBlock).round(3).toFixed(0), 10)
```
clark800: Note that `toFixed` uses the default rounding mode which I believe is set to `ROUND_HALF_EVEN`, so if you want to `floor` it, you would still have to do `round(0, 0)`.

src/engine-actions/translate-swap.js
clark800: I think exposing the error types through an export seems reasonable, and they could all use the same name, like `PermanentError`, so that they can be imported easily. Another option is to use `error.constructor.name` to test which would avoid needing to export them, but I don't see much downside in exporting.

src/engine-actions/translate-swap.js
```diff
@@ -1,227 +1,80 @@
...
+
+  // Translate seconds into blocks, rounding up so that we're sure the end
+  // node will accept the payment. The LND api expects this to be an int32,
+  // so we convert to a javascript number.
+  const finalCltvDelta = parseInt(Big(finalCltvDeltaSecs).div(this.secondsPerBlock).round(0).toFixed(0), 10)
```
clark800: Also here: comment says rounding up, but rounding mode is `ROUND_HALF_EVEN`.

src/engine-actions/translate-swap.js
dannypaz: Im fine with this

src/engine-actions/translate-swap.js
treygriffith: Once it is rounded, toFixed should have no impact (just chop off the trailing zeros). Unless I'm misunderstanding something.

src/engine-actions/translate-swap.js
treygriffith: I'm looking at rounding modes here: https://mikemcl.github.io/big.js/#rm, and `0` should be `ROUND_DOWN`

PR #232, opened 2019-06-26
Getting preimage for invoices as a request/response

treygriffith:
Previously, we waited for invoices to settle, then returning the preimage.

However, with the structure introduced in https://github.com/sparkswap/broker/pull/547 we no longer need to wait for settlement, as we are directly involved in settling the invoice.

As a result, this method can be used exclusively to get preimages from already settled invoices, so its structure can be greatly simplified and reduced to a simple request/response.

It also matches the `lookupInvoice` implementation with what is exposed in the LND RPC.


src/engine-actions/get-settled-swap-preimage.js
```diff
@@ -1,85 +1,41 @@
...
 
-        if (err) {
-          return reject(err)
-        }
+  this.logger.debug('Invoice has been retrieved: ', { state })
 
-        resolve(response)
-      }
+  if (state !== INVOICE_STATES.SETTLED) {
+    throw new Error(`Cannot retrieve preimage from an invoice in a ${state} state.`)
```
dannypaz: need to update this error for the new usage

src/engine-actions/get-settled-swap-preimage.js
```diff
@@ -1,85 +1,41 @@
 const {
-  lookupInvoice,
-  subscribeInvoices
+  lookupInvoice
 } = require('../lnd-actions')
 
+const INVOICE_STATES = Object.freeze({
```
dannypaz: [nit] add jsdoc

clark800: I think the invoice states from the LND API are integers, I added a file to `constants` here that we could use: https://github.com/sparkswap/lnd-engine/pull/234/files#diff-6ee896676d47ea25651fd79124cce3c4R4

clark800: Oh wait, I just noticed there is a configuration for strings, I'll have to change my PR https://github.com/sparkswap/lnd-engine/blob/master/src/utils/load-proto.js#L21

src/engine-actions/get-settled-swap-preimage.js
treygriffith: What do you mean? I think it's already updated?

dannypaz: I read this wrong. ignore the comment

src/engine-actions/is-invoice-paid.js
```diff
@@ -13,7 +13,7 @@ const {
  */
 async function isInvoicePaid (paymentRequestHash) {
   const { paymentHash } = await decodePaymentRequest(paymentRequestHash, { client: this.client })
-  const { settled } = await lookupInvoice(paymentHash, { client: this.client })
+  const { settled } = await lookupInvoice({ rHashStr: paymentHash }, { client: this.client })
```
clark800: What if we did the conversion to bytes in `decodePaymentRequest`? Ideally it would be nice to have a single hash format everywhere outside the lnd-actions directory.

src/lnd-actions/lookup-invoice.js
```diff
@@ -4,14 +4,14 @@ const { deadline } = require('../grpc-utils')
...
- * @param {string} paymentHash
+ * @param {Object} paymentHash
  * @param {Object} opts
  * @param {LndClient} opts.client
  * @returns {Promise<Object>}
  */
 function lookupInvoice (paymentHash, { client }) {
   return new Promise((resolve, reject) => {
-    client.lookupInvoice({ rHashStr: paymentHash }, { deadline: deadline() }, (err, res) => {
+    client.lookupInvoice(paymentHash, { deadline: deadline() }, (err, res) => {
```
clark800: Might be more clear if we change the name to something like `hashObject` if we change the type

src/engine-actions/is-invoice-paid.js
treygriffith: I changed it to the conversion in this function. I want to avoid having custom logic in the `lnd-actions` as they are really meant to just be for calling RPCs on LND.

clark800:
LGTM 👍

PR #233, opened 2019-06-26
added pretty timestamp to transactions list

dannypaz:
We receive and then return a unix epoch timestamp to the caller of `getChainTransactions` which can make it hard to generate reports from this data.

This PR converts timestamps to a ISO 8601 date that is more human readable and just as easily parse able by our client and other tools


dannypaz:
[Timestamps are incorrect format on wallet history](https://trello.com/c/iTOrT1Qg/1737-timestamps-are-incorrect-format-on-wallet-history)

PR #234, opened 2019-06-26
Add waitForSwapCommitment

clark800:
Add waitForSwapCommitment function to support Hold Invoice transition.


clark800:
[Add a `waitForSwapCommitment` method to the engine](https://trello.com/c/Rgq7Qg8T/1721-add-a-waitforswapcommitment-method-to-the-engine)

clark800:
@treygriffith For the `ACCEPTED` transition, if you are subscribed, you will get a notification for the transition to accepted, but if you are looking up afterwards, the invoice only has the creation time and settlement times in it, so I don't know if you can get the accepted transition time afterwards.

For the HTLC expiry, from the invoice, you can look at the RouteHints to get the channel ID, then use ListChannels to get the active channels, select the one with the appropriate ID, and there is a field for pending_htlcs which has the `expiration_height` field on it.

treygriffith:
> @treygriffith For the `ACCEPTED` transition, if you are subscribed, you will get a notification for the transition to accepted, but if you are looking up afterwards, the invoice only has the creation time and settlement times in it, so I don't know if you can get the accepted transition time afterwards.

I suppose we could use creation time as a conservative estimate of the `ACCEPTED` time.

> For the HTLC expiry, from the invoice, you can look at the RouteHints to get the channel ID

I don't think there is a guarantee that the RouteHints will be populated, nor that it will be used. It's an optional field to provide the sender hints about routes to take.

clark800:
> I don't think there is a guarantee that the RouteHints will be populated, nor that it will be used. It's an optional field to provide the sender hints about routes to take.

I guess we would just need to figure out the channel ID some other way. I haven't seen a way to do this in the API yet, but It looks like the paymentRequest contains both the channel ID and expiries encoded in it, which we could decode: https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#tagged-fields

PR #235, opened 2019-06-27
Return creation date from waitForSwapCommitment

clark800:
Update return value of `waitForSwapCommitment` to just the creation date so that we don't expose the invoice data structure which is lnd-specific.


src/engine-actions/wait-for-swap-commitment.js
```diff
@@ -8,7 +8,7 @@ class CanceledSwapError extends Error {}
  * Waits for a swap to enter the ACCEPTED state and errors if it won't happen
  *
  * @param {Bytes} swapHash - hash for a swap
- * @returns {Invoice} - invoice for swap
+ * @returns {Date} - creation date of the HTLC
```
dannypaz: @treygriffith i'm thinking we would want to return a string instead `.toISOString()`?

treygriffith: I think a date makes sense here - it's a native JS type, so passing it shouldn't be a problem.

src/engine-actions/wait-for-swap-commitment.spec.js
```diff
@@ -21,7 +22,8 @@ describe('wait-for-swap-commitment', () => {
       setTimeout(() => stream.emit('data', openInvoice), 1)
       setTimeout(() => stream.emit('data', openInvoice), 2)
       setTimeout(() => stream.emit('data', acceptedInvoice), 3)
-      expect(await waitForSwapCommitment(hash)).to.be.eql(acceptedInvoice)
+      expect(await waitForSwapCommitment(hash)).to.be.eql(
+        new Date(creationDate * 1000))
```
treygriffith: [nit] would be nice to use real dates here rather than just replicating the implementation in the method.

clark800: @treygriffith updated

PR #236, opened 2019-06-27
Add cancelSwap engine action

clark800:
Add cancelSwap engine action in support of hold invoices.


src/engine-actions/cancel-swap.js
```diff
@@ -0,0 +1,13 @@
+const { cancelInvoice } = require('../lnd-actions')
+
+/**
+ * Cancels the invoice for a swap
+ *
+ * @param {string} swapHash - hash of the sawp
```
treygriffith: misspelled swap. Also, is this supposed to be a Base64 string?

src/lnd-actions/cancel-invoice.js
```diff
@@ -0,0 +1,20 @@
...
+ * Cancel a hold invoice
+ *
+ * @param {string} paymentHash
+ * @param {Object} opts
+ * @param {LndClient} opts.client
+ * @returns {Promise<Object>}
+ */
+function cancelInvoice (paymentHash, { client }) {
+  return new Promise((resolve, reject) => {
+    client.cancelInvoice({ paymentHash }, { deadline: deadline() }, (err, res) => {
```
bradleystachurski: Does this need to be `client.invoices.cancelInvoice`?

clark800: Yeah, thanks!

src/engine-actions/cancel-swap.js
clark800: I guess because of the implicit conversion you mentioned it could probably work as either Bytes/Buffer or base64 string but we should try to choose one convention, which I believe is the Bytes/Buffer one? There are other places in the codebase that use `string` so we could update those later.

PR #237, opened 2019-06-28
Feature - Track payment

bradleystachurski:
This PR uses the new `TrackPayment` RPC from LND. This is useful for tracking the status of an in-flight payment, which allows us to make `translateSwap` idempotent. With these changes, we can retire our implementation of `getPaymentPreimage`.


bradleystachurski:
[Use TrackPayment to get payment status](https://trello.com/c/20T1dIfx/1628-use-trackpayment-to-get-payment-status)

src/engine-actions/get-payment-preimage.js
```diff
@@ -27,62 +20,82 @@ const POLLING_INTERVAL = 10000
...
+  return new Promise(async (resolve, reject) => {
+    try {
+      if (!paymentHash) {
+        throw new Error('paymentHash must be defined')
+      }
 
-  if (status === PAYMENT_STATUSES.GROUNDED) {
-    return { permanentError: `No payment with hash '${paymentHash}' is in-flight or complete` }
-  }
+      // TODO: look into how gRPC wants `bytes` in JS
```
treygriffith: our [settings](https://github.com/sparkswap/lnd-engine/blob/master/src/utils/load-proto.js#L20) have `bytes` as Base64 strings.

src/engine-actions/get-payment-preimage.js
```diff
@@ -27,62 +20,82 @@ const POLLING_INTERVAL = 10000
...
-  const payment = payments.find(({ paymentHash }) => {
-    // `paymentHash` returned from ListPayments {@link https://api.lightning.community/#payment}
-    // is a hex string. We need to compare to our payment hash which is a base64 string,
-    // so we convert both to buffers to do the comparison.
-    return paymentHashBuf.equals(Buffer.from(paymentHash, 'hex'))
-  })
+        if (state === PAYMENT_STATUSES.IN_FLIGHT) {
+          // trackPayment will continue to stream updates of the payment state, so
+          // no action is required on our end
+          this.logger.debug('Payment is In-Flight for paymentHash:', { paymentHash })
```
treygriffith: you're missing a `return` here - as is it will fall through to the `throw`

src/engine-actions/get-payment-preimage.js
```diff
@@ -27,62 +20,82 @@ const POLLING_INTERVAL = 10000
...
-  // like the `paymentHash`, the `paymentPreimage` is returned as a hex string from LND,
-  // so we convert to Base64 before returning
-  return Buffer.from(payment.paymentPreimage, 'hex').toString('base64')
+        if (state === PAYMENT_STATUSES.FAILED_TIMEOUT || state === PAYMENT_STATUSES.FAILED_NO_ROUTE) {
+          this.logger.error('Payment has failed', { paymentHash })
+          const permanentError = new Error(`Payment failed for paymentHash: ${paymentHash}`)
+          return finish(null, { permanentError })
+        }
+
+        throw new Error(`Unknown payment status ${state} for payment with hash '${paymentHash}'`)
```
treygriffith: you probably want to `finish` with an error rather than `throw` here

src/engine-actions/get-payment-preimage.js
```diff
@@ -27,62 +20,82 @@ const POLLING_INTERVAL = 10000
...
+      }
+
+      const finish = (err, response) => {
+        stream.removeListener('error', errorListener)
```
treygriffith: `stream.removeAllListeners()` would work here instead of going one by one. It might also let you `reject` or `resolve` in the specific handlers instead of defining a separate `finish` function.

src/engine-actions/get-payment-preimage.js
```diff
@@ -27,62 +20,82 @@ const POLLING_INTERVAL = 10000
...
-  }
+      const stream = await trackPayment(paymentHashBuf, { client: this.client })
 
-  throw new Error(`Unknown payment status ${status} for payment with hash '${paymentHash}'`)
-}
+      const errorListener = (err) => {
+        // TODO: handle gRPC NOT FOUND errors
+        //  It may be best to handle it as a `permanentError`, so we don't have to
+        //  rely on checking NOT FOUND in downstream logic
+        this.logger.error(`Error from trackPayment stream`)
```
treygriffith: Nice to have identifying information in this logs, you can pass an object at the end to add info.

src/engine-actions/translate-swap.js
```diff
@@ -38,40 +39,56 @@ async function translateSwap (takerAddress, swapHash, amount, maxTime, finalCltv
   this.logger.info(`Translating swap for ${swapHash} by sending to ${takerAddress}`,
     { amount, maxTime })
 
-  const { publicKey } = networkAddressFormatter.parse(takerAddress)
+  // To ensure translateSwap is idempotent, we first make a call to getPaymentPreimage to make sure we don't have an
+  // existing payment that is either In-Flight, Completed, or Failed. If there is not an existing preimage for the
+  // provided swapHash, we can be confident that a call to sendPayment will result in creating only one downstream
+  // payment for this swap
+  try {
+    const { paymentPreimage, permanentError } = await getPaymentPreimage(swapHash)
```
treygriffith: might be worth having this return the preimage and `throw` to match the style defined in this file.

src/engine-actions/get-payment-preimage.js
```diff
@@ -27,62 +20,82 @@ const POLLING_INTERVAL = 10000
  * then return the hash.
  *
  * @param  {string} paymentHash - Base64 encoded payment hash
- * @returns {GetPaymentPreimageOutcome}
+ * @returns {Promise<GetPaymentPreimageOutcome>}
  * @throws {Error} If payment is in an unknown status
  */
 async function getPaymentPreimage (paymentHash) {
```
treygriffith: This method may be unnecessary with the changes to `translateSwap`, so you should consider if you can consolidate this into the other method (or at least the other file).

src/lnd-actions/track-payment.js
```diff
@@ -0,0 +1,34 @@
...
+const PAYMENT_STATUSES = Object.freeze({
+  IN_FLIGHT: 'IN_FLIGHT',
+  SUCCEEDED: 'SUCCEEDED',
+  FAILED_TIMEOUT: 'FAILED_TIMEOUT',
+  FAILED_NO_ROUTE: 'FAILED_NO_ROUTE'
+})
+
+/**
+ * Tracks an existing payment
+ * @param {Buffer} paymentHash - hash of payment to track
```
treygriffith: Our settings use `bytes` as a Base64 string.

src/engine-actions/get-payment-preimage.js
bradleystachurski: Perfect, thank you.

src/engine-actions/get-payment-preimage.js
bradleystachurski: Good catch.

src/engine-actions/get-payment-preimage.js
bradleystachurski: Let me know what you think of the style with `finish` removed. Redundant calls to `stream.removeAllListeners()` isn't my favorite, but no biggie either way.

src/engine-actions/translate-swap.js
```diff
@@ -38,40 +50,139 @@ async function translateSwap (takerAddress, swapHash, amount, maxTime, finalCltv
...
+          this.logger.debug('Payment has Succeeded for paymentHash', { paymentHash })
+          stream.removeAllListeners()
+          return resolve(preimage)
+        }
+
+        if (state === PAYMENT_STATUSES.FAILED_TIMEOUT || state === PAYMENT_STATUSES.FAILED_NO_ROUTE) {
+          this.logger.error('Payment failed', { paymentHash })
+
+          stream.removeAllListeners()
+          return reject(new Error(`Payment failed for paymentHash: ${paymentHash}`))
```
treygriffith: shouldn't this also be a `PermanentSwapError`?

bradleystachurski: Yup, good catch.

src/engine-actions/translate-swap.js
```diff
@@ -38,40 +50,139 @@ async function translateSwap (takerAddress, swapHash, amount, maxTime, finalCltv
...
+    // we await so rejections are handled within this function
+    const paymentPreimage = await getPreimageIfPaymentExists(swapHash)
+
+    return paymentPreimage
+  } catch (e) {
+    // If our call to LND returned a gRPC NOT FOUND error, we can be confident there
+    // isn't an existing payment for the provided swapHash, so we can create a new
+    // payment with a call to sendPayment
+    if (e.code !== grpc.status.NOT_FOUND) {
+      throw new PermanentSwapError(e)
```
treygriffith: How did you decide to make this a permanent swap error? It feels like we're in a bit of an unknown state if this call failed (without a NOT FOUND).

treygriffith: Especially since you explicitly mark errors as permanent or not in `getPreimageIfPaymentExists` if feels odd to re-categorize it here.

src/engine-actions/translate-swap.js
```diff
@@ -38,40 +50,141 @@ async function translateSwap (takerAddress, swapHash, amount, maxTime, finalCltv
...
+ * Retrieve the preimage for the given payment hash. If no payment is found
+ * for the hash, it will throw a gRPC NOT FOUND error. If the payment is
+ * in-flight, it will wait until the payment succeeds or is in a permanent
+ * failure state.
+ * @param {string} paymentHash - Base64 encoded payment hash
+ * @param {LndClient} client
+ * @param {Logger} logger
+ * @returns {Promise<string>} - Base64 encoded string of the preimage
+ */
+async function getPreimageIfPaymentExists (paymentHash, client, logger) {
```
treygriffith: [nit] might be nice to have the second parameter as an object with properties `client`, and `logger`

src/engine-actions/translate-swap.js
```diff
@@ -38,40 +50,141 @@ async function translateSwap (takerAddress, swapHash, amount, maxTime, finalCltv
...
+          logger.debug('Payment has Succeeded for paymentHash', { paymentHash })
+          stream.removeAllListeners()
+          return resolve(preimage)
+        }
+
+        if (state === PAYMENT_STATUSES.FAILED_TIMEOUT || state === PAYMENT_STATUSES.FAILED_NO_ROUTE) {
+          logger.error('Payment failed', { paymentHash })
+
+          stream.removeAllListeners()
+          return reject(new PermanentSwapError(`Payment failed for paymentHash: ${paymentHash}`))
```
treygriffith: might be nice to include some info about the failure (`TIMEOUT` or `NO_ROUTE` at a minimum, more info if LND provides it)

src/engine-actions/translate-swap.spec.js
```diff
@@ -57,47 +64,436 @@ describe('translate-swap', () => {
...
+      beforeEach(() => {
+        const error = new Error()
+        error.code = 5 // gRPC NOT FOUND status code
+        getPreimageIfPaymentExists.rejects(error)
+      })
+
+      it('throws an error for any error other than NOT FOUND', () => {
+        getPreimageIfPaymentExists.rejects()
+
+        return expect(translateSwap.call(engine, address, swapHash, amount, maxTime)).to.eventually.be.rejectedWith(Error)
```
treygriffith: Might be better to test the error message itself since almost anything we throw will be an `Error`

treygriffith: Might also be nice to test the pass through of Error types (`PermanentSwapError` and otherwise)

src/engine-actions/translate-swap.spec.js
```diff
@@ -57,47 +64,436 @@ describe('translate-swap', () => {
...
+          } catch (e) {
+            const listenerCount = getAllListernersCount()
+            expect(listenerCount).to.be.eql(0)
+          }
+        })
+      })
+    })
+  })
+
+  describe('integration tests', () => {
```
treygriffith: These aren't quite integration tests... another name maybe?

src/engine-actions/translate-swap.spec.js
```diff
@@ -57,47 +64,436 @@ describe('translate-swap', () => {
...
+
+    it('returns the preimage if the payment exists', async () => {
+      setTimeout(() => stream.emit('data', succeededPayment), 1)
+      const result = await translateSwap.call(engine, address, swapHash, amount, maxTime)
+
+      expect(result).to.be.eql(succeededPayment.preimage)
+    })
+
+    it('throws a permanent error if the payment exists and failed from timeout', () => {
+      setTimeout(() => stream.emit('data', failedTimeoutPayment), 1)
```
treygriffith: [nit] wouldn't `process.nextTick` be equivalent here?

src/engine-actions/translate-swap.spec.js
```diff
@@ -57,47 +64,436 @@ describe('translate-swap', () => {
...
+
+      return expect(translateSwap.call(engine, address, swapHash, amount, maxTime))
+        .to.eventually.be.rejectedWith(`Payment failed for paymentHash`)
+    })
+
+    it('throws a permanent error if the payment exists and failed from no route', () => {
+      setTimeout(() => stream.emit('data', failedNoRoutePayment), 1)
+
+      return expect(translateSwap.call(engine, address, swapHash, amount, maxTime))
```
treygriffith: should test for the `PermanentSwapError` type

src/engine-actions/translate-swap.spec.js
```diff
@@ -57,47 +64,436 @@ describe('translate-swap', () => {
...
+        .to.eventually.be.rejectedWith('Unknown payment status')
+    })
+
+    it('calls sendPayment for gRPC NOT FOUND errors', async () => {
+      const error = new Error()
+      error.code = 5 // gRPC NOT FOUND
+      setTimeout(() => stream.emit('error', error), 1)
+
+      await translateSwap.call(engine, address, swapHash, amount, maxTime)
+      expect(sendPaymentStub).to.have.been.calledOnce()
```
treygriffith: this test might be better as a true integration - testing that it returns the preimage if the downstream payment has not been started yet.

src/engine-actions/translate-swap.js
bradleystachurski: LND will only provide `PaymentState` on failures, so I'll add the `state` to logging.

PR #239, opened 2019-07-02
Setup typescript and apply typescript-related fixes

clark800:
Mostly JSDoc updates for compatibility with typescript


src/engine-actions/cancel-swap.js
```diff
@@ -3,8 +3,8 @@ const { cancelInvoice } = require('../lnd-actions')
 /**
  * Cancels the invoice for a swap
  *
- * @param {Bytes} swapHash - hash of the swap
- * @returns {Promise}
+ * @param {string} swapHash - hash of the swap
+ * @returns {Promise<Object>}
```
zallarak: can `Object` be more specific here?

src/engine-actions/create-refund-invoice.js
```diff
@@ -7,21 +7,21 @@ const { addInvoice, decodePaymentRequest } = require('../lnd-actions')
  * @type {String}
  * @default
  */
-const DEFAULT_INVOICE_EXPIRY = 31536000
+const DEFAULT_INVOICE_EXPIRY = '31536000'
```
zallarak: This seems like a big change. Just flagging to ensure you checked that the value is consumed as a string

src/engine-actions/get-chain-transactions.js
```diff
@@ -22,7 +22,7 @@ const TRANSACTION_TYPES = Object.freeze({
 /**
  * Gets all opening and closing transaction ids from a list of channels.
  * @param {Object} channels
- * @returns {Set} channelTxIds
+ * @returns {Promise<Set>} channelTxIds
```
zallarak: Can you do Set<string>?

src/engine-actions/translate-swap.js
```diff
@@ -63,7 +63,7 @@ async function translateSwap (takerAddress, swapHash, amount, maxTime, finalCltv
     amt: amount,
     finalCltvDelta,
     cltvLimit,
-    feeLimit: DEFAULT_FEE_LIMIT
+    feeLimit: DEFAULT_FEE_LIMIT.toString()
```
zallarak: This seems to be now a string, so is `toString()` does not seem necessary. Also, same comment as the prior variable that was changed to a string (just ensure that it is truly what you wanted)

src/lnd-actions/connect-peer.js
```diff
@@ -5,22 +5,26 @@ const { deadline } = require('../grpc-utils')
...
  * @private
- * @param {Error} err
+ * @param {Object} err
+ * @param {number} err.code
+ * @param {string} err.details
  * @returns {boolean}
  */
 function alreadyConnected (err) {
-  return (err && err.code === 2 && err.details && err.details.includes('already connected to peer'))
+  return (err && err.code === 2 && err.details !== undefined && err.details.includes('already connected to peer'))
```
zallarak: a style guide I read a long time ago suggested doing nullish checks in javascript with `!= null`. Here, if `err.details = null`, then `(err.details !== undefined) === true`.

zallarak: unless you are certain `err.details` will never be `null`

zallarak:
Overall looks solid. Will be a big win!

A nice-to-have would be giving some of the `Object` types more concrete types, either with literals `{ a: number, b: string}`, or interfaces.

Will let someone more familiar with the code give the final ok, unless you'd like me to take a deeper pass.

src/engine-actions/create-refund-invoice.js
clark800: Yeah, the `expiry` of an invoice is an `int64` type in the `rpc.proto` so we would have to use a string to support the full range.

src/lnd-actions/connect-peer.js
clark800: Hmm, I tried that yesterday and I think it gave me a lint error but now it seems to be ok

src/engine-actions/cancel-swap.js
clark800: Yeah, but I'll leave refinements for later; there are tons of things that would be nice to refine.

.gitignore
```diff
@@ -59,3 +59,6 @@ typings/
 
 # docs
 docs/
+
+# temp files
+*.swp
```
treygriffith: Probably belongs in a global ignore?

src/engine-actions/create-channels.js
```diff
@@ -15,10 +15,10 @@ const { CHANNEL_ROUNDING } = require('../constants')
  * uneconomic channels (that are too small to be worth it.)
  *
  * @private
- * @param   {Engine} engine
+ * @param   {Object} engine
```
treygriffith: Is there a way for us to define a global type like `Engine`?

src/engine-actions/create-wallet.js
```diff
@@ -7,7 +7,8 @@ const {
  * Creates a wallet
  *
  * @param {string} password - wallet password, used to unlock lnd wallet
- * @returns {Array<string>} 24 word cipher seed mnemonic
+ * @returns {Promise<Array<string>>} 24 word cipher seed mnemonic
+ * @this {{walletUnlocker: import('../lnd-setup').LndWalletUnlockerClient}}
```
treygriffith: Is this the best way to do this?

src/engine-actions/pay-invoice.js
```diff
@@ -6,12 +6,11 @@ const {
...
  * @param {String} paymentRequest
- * @param {Object} options
- * @param {Number} expiry expiration of refund invoices
- * @return {String} paymentPreimage
+ * @return {Promise<string>} paymentPreimage
  */
 
-async function payInvoice (paymentRequest, options = {}) {
+async function payInvoice (paymentRequest) {
+  // @ts-ignore
```
treygriffith: why the ignore here?

src/engine-actions/create-wallet.js
clark800: I replaced all the inline imports with typedefs to make it cleaner

src/engine-actions/create-channels.js
clark800: Yup, I added a global typedef for this

src/engine-actions/pay-invoice.js
clark800: It appears to be an invalid usage of `sendPaymentSync` according to the LND documentation, though I'm guessing the documentation is just wrong.

treygriffith: I think we should mark this as a TODO to make the parameters for `sendPayment` actually reflect what's required (using either the payment request: https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#bolt-11-invoice-protocol-for-lightning-payments or the actual fields)

clark800: Where should we mark it? Should I make a ticket for enabling strict mode and add this as a note for it?

clark800: I fixed it so that we no longer need the ignore.

PR #240, opened 2019-07-08
Add settleSwap engine action

clark800:
Add settleSwap engine action in support of hold invoices.


src/engine-actions/settle-swap.js
```diff
@@ -0,0 +1,13 @@
+const { settleInvoice } = require('../lnd-actions')
+
+/**
+* Settles the invoice for a swap
+*
+* @param {string} swapHash - hash of the swap
+* @returns {Promise}
+*/
+async function settleSwap (swapHash) {
```
treygriffith: should add some tests

clark800: I couldn't think of much to test, but I added one.

PR #242, opened 2019-07-08
Fix - Fee limits for execute and translate swap

dannypaz:
The SendPayment RPC changes to use a `FeeLimit` object instead of a digit. The FeeLimit object allows you to specify either a fixed amount or percentage based fee. For both execute/translate, these fees are 0.

This PR fixes both instances where we use FeeLimit for send payment


bradleystachurski:
For future viewers, it's worth noting that we use `sendPaymentSync`, which is defined in `rpc.proto` and has `FeeLimit` defined as an object, where the new `SendPayment` rpc is defined in the `router.proto` file that defines `FeeLimit` as an `int64`.

PR #246, opened 2019-07-09
Add default backup path to lnd-engine

dannypaz:
Instead of prompting a user to supply a path to a backup file for LND, we are going to make a decision that the user must put the backup file to `/backup/lnd-btc.backup` or `/backup/lnd-ltc.backup` respectively.

This PR adds a `backupFilePath` configuration and changes `backup` to a boolean. The user can call `recoverWallet` with backup to `false` and the recovery method will only try to recover on-chain funds.


src/engine-actions/recover-wallet.js
```diff
@@ -19,42 +21,39 @@ async function isValidSeed (seed) {
 }
 
 /**
- * Creates a wallet
+ * Recover wallet funds (on-chain and channel backups)
```
bradleystachurski: 👍

src/config.json
```diff
@@ -12,7 +12,9 @@
       "//2": "from bolt-0002, max channel size is 2^24 - 1 satoshis",
       "maxChannelBalance": "16777215",
       "//3": "LND specifies max payment size as UInt32 - 1 millisatohis (https://github.com/lightningnetwork/lnd/blob/v0.5.1-beta/rpcserver.go#L49-L58)",
-      "maxPaymentSize": "4294967"
+      "maxPaymentSize": "4294967",
+      "//4": "LND static channel backup file path",
```
bradleystachurski: I'm not familiar with the config file. What's the purpose of prepending `//` on certain properties?

src/engine-actions/recover-wallet.js
```diff
@@ -19,42 +21,39 @@ async function isValidSeed (seed) {
...
  *
  * @param {string} password - wallet password, used to unlock lnd wallet
  * @param {Array<string>} seed - 24 word cipher seed mnemonic
  * @param {Buffer} backup - Buffer of the LND static channel backup file
```
bradleystachurski: Should this be switched to `bool`?

src/config.json
dannypaz: @bradleystachurski its the way we designate comments. When the file is loaded. the property still exists however it was the easiest way for us to document some of the variables

src/engine-actions/recover-wallet.js
```diff
@@ -19,44 +21,39 @@ async function isValidSeed (seed) {
...
 /**
- * Creates a wallet
+ * Recover wallet funds (on-chain and channel backups)
  *
  * @param {string} password - wallet password, used to unlock lnd wallet
  * @param {Array<string>} seed - 24 word cipher seed mnemonic
- * @param {Buffer} backup - Buffer of the LND static channel backup file
- * @param {?number} recoveryWindow - number in blocks for look back
- * @returns {Promise<Array<string>>} 24 word cipher seed mnemonic
+ * @param {boolean} backup - Buffer of the LND static channel backup file
```
treygriffith: This doc is wrong - it's no longer a buffer. Also a name like `useChannelBackup` might be more descriptive for its current usage.

dannypaz: I dig. Good suggestion

PR #259, opened 2019-07-26
Feature - getNetwork engine action

bradleystachurski:
This PR adds a `getNetwork` engine action. This is used for determining what type of network LND is connected to (e.g. mainnet, regtest).


src/engine-actions/get-network.js
```diff
@@ -0,0 +1,14 @@
+const { getInfo } = require('../lnd-actions')
+
+/**
+ * Returns the lnd network (e.g. mainnet, regtest, etc)
+ *
+ * @returns {Promise<string>} network
+ */
+async function getNetwork () {
+  const { chains } = await getInfo({ client: this.client })
+  const { network } = chains[0]
```
clark800: What if there is more than one chain?

bradleystachurski: This would break, however `lnd-engine` assumes one LND instance per chain so it seems like a safe assumption.

Do you think there's a better way to handle?

clark800: Given that assumption, it sounds like there could still be multiple chains? If so, then the interface might have to take a parameter that uniquely specifys one chain.

treygriffith:
This seems like an unnecessary call - we validate the `chainName` based on the config passed to lnd-engine, so any subsequent check of the chain should be possible just based on the static config.

bradleystachurski:
Closing in favor of getting the network by using `NODE_ENV`.

PR #260, opened 2019-08-03
Mark an invoice as cancelled when it is expired

treygriffith:
This changes the behavior of `waitForSwapCommitment` to treat expired invoices (those past their `expiry` property) the same as invoices that have been cancelled.

To do so, it will reject immediately if an invoice's expiry is past, and it will set a timer to resolve as cancelled for the expiration date.

If `waitForSwapCommitment` resolves in some other fashion before the expiration, the behavior will be unaffected.


treygriffith:
[waitForSwapCommitment should return Canceled if the invoice is expired](https://trello.com/c/vv3D0Elv/86-waitforswapcommitment-should-return-canceled-if-the-invoice-is-expired)

src/engine-actions/wait-for-swap-commitment.js
```diff
@@ -12,25 +12,50 @@ class CanceledSwapError extends Error {}
...
+    })
     stream.on('data', function (invoice) {
       switch (invoice.state) {
         case INVOICE_STATES.OPEN:
+          const creationDate = new Date(invoice.creationDate * 1000)
+          const expirationDate = new Date(
+            creationDate.getTime() + (invoice.expiry * 1000)
+          )
+
+          // if the invoice is open but still expired, we consider it canceled
```
clark800: So LND still calls an invoice "open" after expiration?

src/engine-actions/wait-for-swap-commitment.js
```diff
@@ -12,25 +12,50 @@ class CanceledSwapError extends Error {}
...
+          if (new Date() > expirationDate) {
+            reject(new CanceledSwapError(swapHash))
+            stream.cancel()
+            return
+          }
+          // once the invoice is expired, treat it as cancelled
```
clark800: I'm not clear on the motivation for this change; naively it seems like it would be better to have different errors for cancellation vs expiration in case the caller wants to know.

src/engine-actions/wait-for-swap-commitment.js
clark800: Should "still expired" be "already expired"?

src/engine-actions/wait-for-swap-commitment.js
```diff
@@ -12,25 +12,50 @@ class CanceledSwapError extends Error {}
  */
 async function waitForSwapCommitment (swapHash) {
   return new Promise((resolve, reject) => {
+    // we use an array for timers since an invoice can be
+    // trigger the stream in an OPEN state multiple times.
```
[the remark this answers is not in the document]
treygriffith: LND will trigger this stream for any update to the invoice, not necessarily just the state, and we have a test case reflecting that. I'm not sure exactly what properties change that trigger it.

src/engine-actions/wait-for-swap-commitment.js
treygriffith: Perhaps, but how would you handle them differently? They are both cases where the invoice is not and will never be settled based on the action of this actor (either explicitly cancelling, or by implicitly cancelling through expiration).

src/engine-actions/wait-for-swap-commitment.js
treygriffith: Yes, the state of the invoice is managed separately from whether it is expired. Cancellation is one reason an invoice will be rejected, expiration is another.

src/engine-actions/wait-for-swap-commitment.js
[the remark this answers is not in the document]
clark800: Ok, do we actually need more than one timer or could we just check to see if a timer already exists each time?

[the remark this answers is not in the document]
treygriffith: Yep good call

PR #261, opened 2019-08-03
add an initiateSwap action that takes explicit time locks

treygriffith:
This change adds a new action, `initiateSwap`, that is like `executeSwap`, except that it does not specify the time locks internally.

This is similar to the change made previously to `prepareSwap` in #229


treygriffith:
[Fix timelock calculations for execute swap](https://trello.com/c/5Qe7nrde/97-fix-timelock-calculations-for-execute-swap)

src/engine-actions/initiate-swap.js
```diff
@@ -0,0 +1,55 @@
...
+async function initiateSwap (address, swapHash, amount, maxTimeLock, finalDelta, feeLimit = DEFAULT_FEE_LIMIT) {
+  this.logger.info(`Initiating swap for ${swapHash} with ${address}`, { address, swapHash, amount })
+
+  const { publicKey: counterpartyPubKey } = networkAddressFormatter.parse(address)
+
+  // Since `maxTimeLock` is a maximum, we use Math.floor to be conservative
+  const cltvLimit = Math.floor(maxTimeLock / this.secondsPerBlock)
+
+  // We use Math.ceil to be sure that the last node in this leg of the swap accepts the payment.
+  const finalCltvDelta = finalDelta ? Math.ceil(finalDelta / this.secondsPerBlock) : 0
```
clark800: Is the ternary necessary here?

src/engine-actions/initiate-swap.spec.js
```diff
@@ -0,0 +1,95 @@
...
+      client,
+      logger: {
+        error: sinon.stub(),
+        info: sinon.stub()
+      },
+      secondsPerBlock: 600
+    }
+  })
+
+  it('sends a payment', async () => {
```
clark800: Perhaps a matter of opinion, but I think this might be simpler if all the positive test cases were consolidated into a single test case.

src/engine-actions/initiate-swap.js
treygriffith: If it's not specified it should use the LND default

clark800: The JSDoc says it's a required parameter and typescript should enforce it, though maybe we haven't tested JSDoc typescript across projects, would be a good thing to verify at some point.

PR #263, opened 2019-08-05
fix lookup invoice on prepare swap

treygriffith:
Prepare swap was using an old signature for `lookupInvoice`. This change fixes that bug.


treygriffith:
[Retry executing quotes on restart](https://trello.com/c/nHCfP1UI/20-retry-executing-quotes-on-restart)

PR #265, opened 2019-08-06
Fix: pass hold invoice params directly

treygriffith:
This fixes a bug in `addHoldInvoice` where we were destructuring a subset of params, and inadvertently stripping one out (`cltvExpiry`). This fix changes that by removing the destructuring and passing through directly.


treygriffith:
[Retry executing quotes on restart](https://trello.com/c/nHCfP1UI/20-retry-executing-quotes-on-restart)

PR #269, opened 2019-08-08
Handle cancel events when waiting for swaps

treygriffith:
This change handles the grpc `CANCELLED` error, which is emitted when calling `.cancel()` on a readable stream.


treygriffith:
[Unhandled CANCELLED error if swap is expired](https://trello.com/c/vVSHDYNb/139-unhandled-cancelled-error-if-swap-is-expired)

PR #273, opened 2019-09-03
Add LND version validation

clark800:


src/engine-actions/get-status.spec.js
```diff
@@ -18,12 +18,14 @@ describe('get-status', () => {
     beforeEach(() => {
       getInfoResponse = {
         chains: [{ chain: 'bitcoin', network: 'testnet' }],
+        version: '0.7.1',
```
treygriffith: is this the format that LND returns? I would think it would have `-beta` in it

src/engine-actions/get-status.js
```diff
@@ -21,6 +21,20 @@ const {
  */
 const WALLET_EXISTS_ERROR_MESSAGE = 'wallet already exists'
 
+function parseVersion (version) {
```
treygriffith: does this handle (and does it need to handle) versions that start with `v`

src/engine-actions/get-status.js
```diff
@@ -21,6 +21,20 @@ const {
...
+  const dottedVersion = version.split(' ')[0].split('+')[0].split('-')[0]
+  return dottedVersion.split('.').map(x => parseInt(x, 10))
+}
+
+function compareVersions (a, b) {
+  return b.length === 0 || (a.length > 0 &&
+    (a[0] > b[0] || (a[0] === b[0] && compareVersions(a.slice(1), b.slice(1)))))
+}
+
+function checkVersion (version, minVersion) {
```
treygriffith: something a bit non-intuitive about this function is that `checkVersion('0.7.1-beta-rc1', '0.7.1-beta-rc2')` returns `true`.

src/engine-actions/get-status.spec.js
clark800: Yeah, it currently returns beta but the comparison function strips that part off. The stripping is tested in other unit tests.

src/engine-actions/get-status.js
clark800: It doesn't handle that and lnd doesn't currently return a `v` so shouldn't be necessary.

src/engine-actions/get-status.js
clark800: Yeah, parsing those parts would make it a lot more complicated. The most popular packages on npm don't handle this properly either:

https://www.npmjs.com/package/node-version-compare
(https://github.com/leohihimax/node-version-compare/blob/master/index.js)
https://www.npmjs.com/package/compare-versions

treygriffith: Probably worth making a note of that somewhere. To be extra safe, we could disallow the `minVersion` in lnd engine from having anything after the dash.

https://www.npmjs.com/package/compare-versions handles it properly from what I can tell: https://runkit.com/embed/91p1gatstxfx

clark800: I checked the semver docs and it looks like `compare-versions` does handle semver properly, but semver orders fields with letters lexicographically so for example both of these return the same result even though we'd expect them to be different (because semver would probably use a dot between `rc` and the number).

```
var compareVersions = require("compare-versions")
console.log(compareVersions('v0.1.0-beta-rc2', 'v0.1.0-beta-rc1'))
console.log(compareVersions('v0.1.0-beta-rc2', 'v0.1.0-beta-rc11'))
```

Are you ok with using the library? (I know you typically want to avoid more dependencies)

treygriffith: I think given that LND follows semver, and semver is a bit tricky, and this library itself has no dependencies, this is a case where it may make sense to use a library.

PR #275, opened 2019-09-06
Specify channel target confirmations

treygriffith:
This change lets you take advantage of LND's `targetConf` to specify the number of blocks to target when determining the channel opening fee. This allows more fine-grained control over channel-opening behavior.

It also exposes the `private` parameter to make channels private.

It is a **breaking change** for users of the `roundBehavior` parameter of `createChannels`, which was replaced with an options object.


treygriffith:
[Increase channel opening fees on server](https://trello.com/c/TOPjHN2v/273-increase-channel-opening-fees-on-server)

src/engine-actions/create-channels.js
```diff
@@ -140,14 +146,22 @@ async function assertBalanceIsSufficient (engine, fundingAmount) {
  *
  * @param {string} paymentChannelNetworkAddress
  * @param {string} fundingAmount - int64 string
- * @param {string} roundBehavior - Behavior when encountering a channel that is too small
+ * @param {object} [options]
+ * @param {string} [options.roundBehavior=CHANNEL_ROUNDING.DOWN] - Behavior when encountering a channel that is too small
+ * @param {number} [options.targetTime=DEFAULT_CONFIRMATION_DELAY] - Estimated time to first confirmation in seconds
```
clark800: This could be pretty confusing if you don't have the context of fees, `confirmationTargetTimeForFeeRate` is my attempt to capture it in a name, but just adding something about fee targeting to the comment could be good

PR #281, opened 2019-09-18
Add time lock fields to engine

clark800:
Also remove some stale code (isPaymentPendingOrComplete)


src/index.js
```diff
@@ -83,6 +84,9 @@ class LndEngine {
       this[configKey] = this.currencyConfig[configKey]
     })
 
+    this.outboundTimeLock = outboundTimeLock || 9 * this.secondsPerBlock
```
dannypaz: sorry, this is a [nit]. Would be worth to have some kind of documentation to show where `9` and `144` came from

src/index.js
```diff
@@ -69,6 +79,7 @@ class LndEngine {
     this.symbol = symbol
     this.minVersion = minVersion
     this.currencyConfig = currencies.find(({ symbol }) => symbol === this.symbol)
+    this.secondsPerBlock = 600 // can be overridden by config.json setting
```
treygriffith: why have this default in here? Seems like it's confusing more than anything

src/index.js
```diff
@@ -83,6 +94,17 @@ class LndEngine {
       this[configKey] = this.currencyConfig[configKey]
     })
 
+    // 90 minutes, from BOLT #2
+    // see https://github.com/lightningnetwork/lightning-rfc/blob/master/
+    //     02-peer-protocol.md#cltv_expiry_delta-selection
+    this.outboundTimeLock = outboundTimeLock || 9 * this.secondsPerBlock
```
treygriffith: should this be 9 blocks, or 90 minutes? (LTC will have a different block time)

src/index.js
```diff
@@ -83,6 +94,17 @@ class LndEngine {
...
+
+    // based on a forward time lock delta of 40 blocks from LND
+    // see https://github.com/lightningnetwork/lnd/pull/2759
+    this.retrieveWindowDuration =
+      retrieveWindowDuration || 10 * this.secondsPerBlock
```
treygriffith: same comment - may make sense to define this in seconds rather than blocks given that this engine supports two blockchains

src/index.js
clark800: It's for the type checker. We could also do a ts-ignore comment but was trying to avoid it.

src/index.js
clark800: The formula in the link in the comment calculates the value in terms of blocks so was following that.

src/index.js
clark800: I don't know if there is a formula for this one, but it kind of makes sense that blockchain congestion would be relieved in proportion to block frequency.

src/index.js
treygriffith: The link in the comment is for Bitcoin only

src/index.js
treygriffith: perhaps for publish duration, but for retrieval I'd argue that it's more likely about network connectivity or downtime.

src/index.js
treygriffith: hmm, feels a little bit weird, but I suppose it's ok.

src/index.js
clark800: Yeah, so do you think there is a better approach? The calculation there seems like it could make sense for other blockchains, though the "optimal" value would probably depend on other factors like network difficulty, requiring a custom value for each blockchain.

src/index.js
clark800: Ok I think that makes sense, I updated the retrieveWindowDuration to be independent of block time

PR #283, opened 2019-10-02
Remove validateEngine

clark800:
Previously the engine would continually retry validation even when the LND connection configuration was incorrect; now it only retries when LND is in a NOT_SYNCED state.


dannypaz:
im cool with this. @clark800 did we want to remove `validateEngine` entirely since we are no longer using on [sparkswap desktop](https://github.com/sparkswap/sparkswap-desktop)

clark800:
@dannypaz yeah, I just updated to remove validateEngine

PR #285, opened 2019-10-04
remove execute swap and add block buffer to the engine

treygriffith:
This change removes `executeSwap` (which was replaced by `initiateSwap`) and adds the `blockBuffer` as a property on the engine.


src/engine-actions/index.js
```diff
@@ -65,7 +64,6 @@ module.exports = {
     isBalanceSufficient,
     payInvoice,
     prepareSwap,
-    executeSwap,
```
clark800: So is `settleSwap` replacing `executeSwap` now? I'd be interested in learning the story behind this.

src/index.js
```diff
@@ -102,6 +104,13 @@ class LndEngine {
     this.retrieveWindowDuration = retrieveWindowDuration || 6000
     this.claimWindowDuration = claimWindowDuration || 30 * this.secondsPerBlock
 
+    // The amount of time, in seconds, that we'd like to buffer any output timelock
+    // by to account for block ticks during a swap This is especially problematic on
+    // regtest where we mine blocks every 10 seconds and is a known issue on
+    // mainnet.
+    // @see {@link https://github.com/lightningnetwork/lnd/issues/535}
+    this.blockBuffer = blockBuffer || 1200
```
clark800: nit: Since this seems to have more to do with blocks maybe we could express it as `2 * this.secondsPerBlock`

src/index.js
```diff
@@ -102,6 +104,13 @@ class LndEngine {
     this.retrieveWindowDuration = retrieveWindowDuration || 6000
     this.claimWindowDuration = claimWindowDuration || 30 * this.secondsPerBlock
 
+    // The amount of time, in seconds, that we'd like to buffer any output timelock
```
clark800: I think we can add a little more explanation: "The recipient requires a minimum number of blocks for the time lock on an inbound HTLC. If we use this exact value to create the HTLC on the sending side, there is a small chance that a block will be mined during the time it takes for the HTLC to be sent over the network to the recipient, and this will cause the recipient to reject the HTLC."

src/engine-actions/index.js
treygriffith: `initiateSwap` replaced `executeSwap`

PR #287, opened 2019-10-17
added get invoice for lnd-engine

dannypaz:
Adds a `getInvoice` call so we can get destination and value of invoice


dannypaz:
[Set up process for delivering satoshis to users for referral program](https://trello.com/c/BA3nTTRk/484-set-up-process-for-delivering-satoshis-to-users-for-referral-program)

PR #307, opened 2020-02-06
Update proto to 0.8.0 definitions

dannypaz:


dannypaz:
we'll hold off on this change until we require 0.8.0 for clients

PR #308, opened 2020-02-18
Update proto to 0.7.1

dannypaz:
Our current lnd-engine proto files are locked to 0.7.0, but there are a few updates from 0.7.1 (the min version for clients) that we should include

This PR also bumps version to 0.10.0-beta


bradleystachurski:
[Update `lnd-engine` proto](https://trello.com/c/fYFeq2FM/934-update-lnd-engine-proto)

